PR #1, opened 2024-02-05
Fix README.md

ream88:
I'm probably nitpicking here, but the `Endpoint` module is usually created in the `*Web` namespace 😉


germsvel:
Not nitpicking at all. Thanks for catching that and opening a PR! 🙌

PR #2, opened 2024-02-05
Update README.md

srikanthkyatham:
test -> :test for deps part


germsvel:
Thanks @srikanthkyatham. Seems like this got fixed in https://github.com/germsvel/phoenix_test/pull/3. Don't know how I missed your PR. But thanks for opening it! 🤗

PR #8, opened 2024-02-07
Allow piping from refute_has

### LINKED ISSUE #7: `refute_has` can't be used in pipes in the same way `assert_has` can
Chaining assertions with the pipe operator works very well and allows for readable test cases:

```elixir
conn
|> visit("/")
|> assert_has(".status", "disabled")
|> click("Enable")
|> assert_has(".status", "enabled")
```

This works with `assert_has` neatly, but it doesn't work with `refute_has` in the same way one would expect:

```elixir
conn
|> visit("/")
|> refute_has(".status", "enabled")
|> click("Enable")
|> assert_has(".status", "enabled")
```
cashes with
```
     ** (Protocol.UndefinedError) protocol PhoenixTest.Driver not implemented for false of type Atom. This protocol is implemented for the following type(s): PhoenixTest.Live, PhoenixTest.Static
```

soundmonster:
Allow using `refute_has` in pipes in the same way `assert_has` already works:

Fixes #7

### DIFF
--- a/lib/phoenix_test/assertions.ex
+++ b/lib/phoenix_test/assertions.ex
@@ -55,6 +55,8 @@ defmodule PhoenixTest.Assertions do
         But found #{Enum.count(elements)} elements with selector #{inspect(selector)} and text #{inspect(text)}:
         """
     end
+
+    session
   end
 
   defp format_found_elements(elements) do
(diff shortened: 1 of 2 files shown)


germsvel:
Thanks for opening the issue and submitting a fix @soundmonster! 🙌 🙌 🙌

PR #9, opened 2024-02-07
Fix formatting found elements with nested children

soundmonster:
As part of its assertion error messages, PhoenixTest outputs HTML elements that match the query selector but do not
contain the requested text. Currently it only works for elements that only contain one child, and breaks otherwise. It
is sometimes useful to assert that a word will be present in a complex structure (maybe generated, like the modal popup
from `PetalComponents`) without coupling the code too much to the internal DOM of the structure.

Proposal: allow complex structures but not assuming only one child by default, and use `Floki.raw_html(content, pretty: true)` to render the error message.

### DIFF
--- a/lib/phoenix_test/assertions.ex
+++ b/lib/phoenix_test/assertions.ex
@@ -62,6 +62,10 @@ defmodule PhoenixTest.Assertions do
   end
 
   defp format_found_element({tag, _attrs, [content]}) do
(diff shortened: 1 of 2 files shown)


germsvel:
😍 `Floki.raw_html/2`! That's what I was looking for. Thanks for making this change. I think it's a great improvement!

Issue #10, opened 2024-02-07
Documentation link to file on local machine


igor-ladkin:
@germsvel hi!

I found a link in readme that still leads to a file on a local machine.

https://github.com/germsvel/phoenix_test/blob/98cd57ef2316dbe611b1469e5b3a1d3e697f0db4/README.md?plain=1#L129

It should probably be https://hexdocs.pm/phoenix_test/PhoenixTest.html instead.

PR #11, opened 2024-02-07
Use correct URL for module documentation

igor-ladkin:
Fixes https://github.com/germsvel/phoenix_test/issues/10


germsvel:
🎉 thanks for catching that and submitting a fix!

PR #13, opened 2024-02-07
Raise AssertionError instead of RuntimeError

soundmonster:
Currently, assertions `assert_has` and `refute_has` raise `RuntimeError`

```
  1) test index shows empty states (Manage.PageLiveTest)
     apps/manage/test/manage/live/page_test.exs:6
     ** (RuntimeError) Could not find any elements with selector "span#bar".

     code: |> assert_has("span#bar", "foo")
     stacktrace:
       (phoenix_test 0.2.1) lib/phoenix_test/assertions.ex:17: PhoenixTest.Assertions.assert_has/3
       test/manage/live/page_test.exs:9: (test)
```

This works, but it's more common for assertion helpers to raise `ExUnit.AssertionError` instead, as it signals that
there is something wrong with what the test author is asserting vs. something wrong with the setup as such.

This PR introduces `AssertionError` to the `Assertions` module, and only there. This is how output of the same test
looks like:

```
  1) test index shows empty states (Manage.PageLiveTest)
     apps/manage/test/manage/live/page_test.exs:6
     Could not find any elements with selector "span#bar".

     code: |> assert_has("span#bar", "foo")
     stacktrace:
       (phoenix_test 0.2.1) lib/phoenix_test/assertions.ex:18: PhoenixTest.Assertions.assert_has/3
       test/manage/live/page_test.exs:9: (test)
```

### DIFF
--- a/lib/phoenix_test/assertions.ex
+++ b/lib/phoenix_test/assertions.ex
@@ -2,6 +2,7 @@ defmodule PhoenixTest.Assertions do
   @moduledoc false
 
   import ExUnit.Assertions
+  alias ExUnit.AssertionError
 
   alias PhoenixTest.Query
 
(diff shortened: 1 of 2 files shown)


germsvel:
Love it! 🙌 Thanks for doing this.

PR #14, opened 2024-02-08
Use Floki to find nested fields

Nilsonn:
In the current version, `fill_form` can't be used if the fields are no direct children of the form element.
This PR changes the `build_fields` function to use Floki in order to find all input, select and textarea elements that are somewhere in the tree of the form element.


germsvel:
This is a great change. Thanks for making it! 🥳

PR #15, opened 2024-02-08
Support file input with plug upload

Nilsonn:
When using an input field for a file, the `validate_expected_fields` function raises.

This PR solves this issue but includes `Plug.Upload` pattern matches in some functions. I was not quite sure whether this is a good approach but wanted to suggest the solution anyway.


germsvel:
@Nilsonn thanks so much for the PR! I haven't had time to think about how to support file uploads yet 😅 so this is great! 

Before I merge it, I want to think a little about the API for testing uploads since we have to support both LiveView and static pages. Until then, I'm going to hold off on merging this. I want to test a few things and see what would be good. 

And, of course, I'm always open to suggestions. Do you have any thoughts on how to support uploads nicely for both LiveView and static pages?

PR #20, opened 2024-02-10
Allow multiple matching elements in assert_has

### LINKED ISSUE #18: CaseClauseError in `assert_has` when multiple matches
[The docs for assert_has](https://hexdocs.pm/phoenix_test/PhoenixTest.html#assert_has/3) say "It'll raise an error if no elements are found, but it will not raise if more than one matching element is found."

But I'm seeing that an `assert_has` with multiple matching elements is raising a `CaseClauseError` when it gets back a `{:found_many, ...}` tuple.

If the docs are correct, it looks like it should be possible to add another clause to [`Assertions.assert_has`](https://github.com/germsvel/phoenix_test/blob/main/lib/phoenix_test/assertions.ex#L14-L33). Probably something like:

```elixir
{:found_many, _found}
  -> assert true
```

and a test case for this.

`refute_has` does have a clause for `{:found_many, ...}`, so no problems there.

I'll work on a PR to address this.

randycoulman:
Fixes #18

### DIFF
--- a/lib/phoenix_test/assertions.ex
+++ b/lib/phoenix_test/assertions.ex
@@ -15,6 +15,9 @@ defmodule PhoenixTest.Assertions do
       {:found, _found} ->
         assert true
 
+      {:found_many, _found} ->
+        assert true
+
       {:not_found, []} ->
         raise AssertionError,
           message: """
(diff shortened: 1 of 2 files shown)


germsvel:
Thanks so much for this fix @randycoulman! Seems like I missed that clause when updating the assertions to use the `Query` module.

Issue #17, opened 2024-02-09
Document how to `fill_form` for a changeset with associations


mayel:
How should one `fill_form` or `submit_form` when you have inputs with a `name` such as `account[email_address]` or even `account[credential][password]`?

I have tried:
```
    account: [
      email_address: email, 
      credential: [password: password]
    ]
```

```
    %{
        "account" => %{
          "email_address" =>  email,
          "credential" => %{"password" => password}
        }
      }
```

```
    %{
      "account[email_address]" => email,
      "account[credential][password]" => password
    }
```

ending up with this one which seems to work but isn't very ergonomic:
```
      %{
        "name='account[email_address]'" => email,
        "name='account[credential][password]'" => password
      }
```

randycoulman:
@mayel I was looking at this as well. Internally, for LiveView forms, it looks like PhoenixTest is using [Phoenix.LiveViewTest.form/3](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveViewTest.html#form/3) to fill in the form fields, so whatever works there should work here.

I've been able to get a simple form (no nested fields) to work with `account: [name: "Whatever"]`. It seems like your first option above should work, but I haven't tried with nested fields yet.

If you're using a LiveView form, then maybe you could try getting it to work with just the `email_address` field first, then add the nested field after that?

germsvel:
Hi @mayel, thanks so much for opening this issue. Seems like I didn't document that well at all! 😅 

You should be able to pass nested maps as the final argument in `fill_form/3` and `submit_form/3`. The nested maps should match the nesting in your inputs. 

So for an `<input name="account[email_address]">`, you should be able to use something like this: 

```elixir
...
|> fill_form("form", account: %{email_address: "mayel@example.com"})

# also works with submit_form
...
|> submit_form("form", account: %{email_address: "mayel@example.com"})
```

I tried to improve the documentation in https://github.com/germsvel/phoenix_test/commit/680938926b5d6024b0e3061dcff01999c74b2c7b, but if you have thoughts on how I could make it even clearer, I'd love to know (and always happy to review PRs).

mayel:
Thanks both! :) 

@germsvel Unfortunately I get this error when trying that:
```
     ** (ArgumentError) Could not find element with selector "#signup-form [name=account]"
     code: |> fill_form("#signup-form",
     stacktrace:
[...]

PR #25, opened 2024-02-14
Failing test for nested field in liveview

totaltrash:
Hi there, thanks for this library, it is sweet as.

I'm seeing the same thing as #17 and have tracked it down to being an issue for LiveView nested form fields. Hopefully this failing test will highlight what's happening

```
mix test
.........................................................................

  1) test fill_form + click_button nested form (PhoenixTest.LiveTest)
     test/phoenix_test/live_test.exs:171
     ** (ArgumentError) Could not find element with selector "#nested-form [name=user]"
     code: |> fill_form("#nested-form", user: %{email: "some@example.com"})
     stacktrace:
       (phoenix_test 0.2.4) lib/phoenix_test/query.ex:9: PhoenixTest.Query.find!/2
       (elixir 1.16.1) lib/enum.ex:987: Enum."-each/2-lists^foreach/1-0-"/2
       (phoenix_test 0.2.4) lib/phoenix_test/live.ex:120: PhoenixTest.Driver.PhoenixTest.Live.fill_form/3
       test/phoenix_test/live_test.exs:174: (test)

......................
Finished in 0.2 seconds (0.2s async, 0.00s sync)
96 tests, 1 failure
```

### DIFF
--- a/test/phoenix_test/live_test.exs
+++ b/test/phoenix_test/live_test.exs
@@ -168,6 +168,14 @@ defmodule PhoenixTest.LiveTest do
       |> assert_has("#form-data", "email: some@example.com")
     end
 
+    test "nested form", %{conn: conn} do
+      conn
+      |> visit("/live/index")
+      |> fill_form("#nested-form", user: %{email: "some@example.com"})
+      |> click_button("Save nested")
+      |> assert_has("#form-data", "email: some@example.com")
+    end
+
     test "can handle forms with inputs, checkboxes, selects, textboxes", %{conn: conn} do
       conn
       |> visit("/live/index")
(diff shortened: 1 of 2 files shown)


germsvel:
A failing test 🤩 Thanks @totaltrash! That'll be super helpful in fixing the issue. I'm honestly shocked (and a bit ashamed) to see that that test didn't already exist -- and that it fails. In theory, LiveView should handle that form, but I think I know what's happening now.

germsvel:
@totaltrash, commit 180dc0dd7e57427304c9cb144ff0f75f5f119faa fixes the issue. I thought there was something more going on with the failing test, so I didn't tag that as closing this or fixing #17. But it turns out, after that commit, the test was only failing because of a test LiveView implementation when rendering the form fields that were submitted. 

In any case, I didn't know if GitHub allowed me to push to your branch. So instead, I opened #26 to capture the test changes you introduced here. 

Thanks again for the failing test!

PR #31, opened 2024-02-17
Add `open_browser/1`

sodapopcan:
Hey German!

Me again.

First off I'd like to say that I'll be totally cool if you don't think this belongs in PhoenixTest, but I did a quick implementation and wanted to create the PR to get the discussion going.

So this introduces a `preview/1` function which is essentially a wrapper around `Phoenix.LiveViewTest.open_browser/1`.

## The problem with using `open_browser/1` directly

PhoenixTest's functions are all pipeable taking in a "session" as the first arg.  With `open_browser` essentially being `IO.inspect` for LiveViewTest, we want to be able to stick one in the middle of a pipeline.  We can't use `open_browser` for this since it takes a `Phoenix.LiveViewTest.View`.  Instead we have to change:

```elixir
conn
|> click_link("foo")
|> fill_form("form", foo: %{bar: "baz"})
|> click_button("Submit")
|> assert_has("Profit!")
```

into:

```elixir
sess =
  conn
  |> click_link("foo")

open_browser(sess.view)

sess
|> fill_form("form", foo: %{bar: "baz"})
|> click_button("Submit")
|> assert_has("Profit!")
```

and back just to "quickly" look at something in the browser.

I'm advocating for this to be a part of PhoenixTest because even though I just started using it yesterday, I'm already pretty in love with it and would prefer not to carry around my own `preview` function to all my projects (but I will if I have to 🙂).

In a nutshell, `preview` simply does the following:

```elixir
def preview(session) do
  Phoenix.LiveViewTest.open_browser(session.view)

  session
end
```

## Why call it `preview/1`?

Since this function can only work on LiveView tests, I wanted to come up with a more generic name in case some very rough equivalent could be implemented for controller tests.  As it stands, this implementation simply raises if given in a controller test.

**DECISION:** We're gonna call it `open_browser`.

## Problems with this PR

The tests I have now basically just ensure it raises in controller tests but otherwise takes a dummy code path to test the LiveView functionality.  In couldn't come up with a good regression test that was simple and un-intrusive.  I took inspiration from how `open_browser` is tested so that's why `preview/2` exists, but I realized it's really not the same thing.  I think something like Mox would be the proper way to go but I didn't want to introduce that without asking.

Anyway, I'm open to any and all feedback, this is a bit of a draft.  Thanks again for this project!

### DIFF
--- a/test/phoenix_test/live_test.exs
+++ b/test/phoenix_test/live_test.exs
@@ -333,4 +333,21 @@ defmodule PhoenixTest.LiveTest do
       end
     end
   end
+
+  describe "open_browser" do
+    setup do
+      open_fun = fn view ->
+        assert %Phoenix.LiveViewTest.View{} = view
+      end
+
+      %{open_fun: open_fun}
+    end
+
+    test "opens the browser", %{conn: conn, open_fun: open_fun} do
+      conn
+      |> visit("/live/index")
+      |> open_browser(open_fun)
+      |> assert_has("h1", "LiveView main page")
+    end
+  end
 end
--- a/lib/phoenix_test.ex
+++ b/lib/phoenix_test.ex
@@ -172,6 +172,12 @@ defmodule PhoenixTest do
   defdelegate submit_form(session, selector, data), to: Driver
 
   @doc """
+  Open the default browser to display current HTML of `session`.
+  """
+  defdelegate open_browser(session), to: Driver
+  defdelegate open_browser(session, open_fun), to: Driver
+
+  @doc """
   Assert helper to ensure an element with given CSS selector and `text` is
   present.
 
--- a/lib/phoenix_test/driver.ex
+++ b/lib/phoenix_test/driver.ex
@@ -8,4 +8,6 @@ defprotocol PhoenixTest.Driver do
   def click_button(session, selector, text)
   def fill_form(session, selector, form_data)
   def submit_form(session, selector, form_data)
+  def open_browser(session)
+  def open_browser(session, open_fun)
 end
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -156,6 +156,11 @@ defimpl PhoenixTest.Driver, for: PhoenixTest.Live do
     phx_change != nil && phx_change != ""
   end
 
+  def open_browser(%{view: view} = session, open_fun \\ &Phoenix.LiveViewTest.open_browser/1) do
+    open_fun.(view)
+    session
+  end
+
   defp maybe_redirect({:error, {:redirect, %{to: path}}}, session) do
     PhoenixTest.visit(session.conn, path)
   end
--- a/lib/phoenix_test/open_browser.ex
+++ b/lib/phoenix_test/open_browser.ex
@@ -1,0 +1,76 @@
+defmodule PhoenixTest.OpenBrowser do
+  @moduledoc false
+
+  # This module contains private functionality ported over from
+  # `Phoenix.LiveViewTest` to make `open_browser` work with `Static` tests.
+
+  @doc """
+  Fully qualifies static assets paths so the browser can find them.
+  """
+  def prefix_static_paths(node, endpoint) do
+    static_path = static_path(endpoint)
+
+    case node do
+      # Remove script tags
+      {"script", _, _} -> nil
+      # Skip prefixing src attributes on anchor tags
+      {"a", _, _} = link -> link
+      {el, attrs, children} -> {el, maybe_prefix_static_path(attrs, static_path), children}
+      el -> el
+    end
+  end
+
+  defp static_path(endpoint) do
+    static_url = endpoint.config(:static_url) || []
+    priv_dir = :otp_app |> endpoint.config() |> Application.app_dir("priv")
+
+    if Keyword.get(static_url, :path) do
+      priv_dir
+    else
+      Path.join(priv_dir, "static")
+    end
+  end
+
+  defp maybe_prefix_static_path(attrs, nil), do: attrs
+
+  defp maybe_prefix_static_path(attrs, static_path) do
+    Enum.map(attrs, fn
+      {"src", path} -> {"src", prefix_static_path(path, static_path)}
+      {"href", path} -> {"href", prefix_static_path(path, static_path)}
+      attr -> attr
+    end)
+  end
+
+  defp prefix_static_path(<<"//" <> _::binary>> = url, _prefix), do: url
+
+  defp prefix_static_path(<<"/" <> _::binary>> = path, prefix) do
+    "file://#{Path.join([prefix, path])}"
+  end
+
+  defp prefix_static_path(url, _), do: url
+
+  @doc """
+  System agnostic function to open the default browser with the given `path`.
+
+  This is ripped verbatim from `Phoenix.LiveViewTest`.
+  """
+  def open_with_system_cmd(path) do
+    {cmd, args} =
+      case :os.type() do
+        {:win32, _} ->
+          {"cmd", ["/c", "start", path]}
+
+        {:unix, :darwin} ->
+          {"open", [path]}
+
+        {:unix, _} ->
+          if System.find_executable("cmd.exe") do
+            {"cmd.exe", ["/c", "start", path]}
+          else
+            {"xdg-open", [path]}
+          end
+      end
+
+    System.cmd(cmd, args)
+  end
+end
--- a/lib/phoenix_test/static.ex
+++ b/lib/phoenix_test/static.ex
@@ -27,6 +27,7 @@ defimpl PhoenixTest.Driver, for: PhoenixTest.Static do
 
   alias PhoenixTest.Html
   alias PhoenixTest.Query
+  alias PhoenixTest.OpenBrowser
 
   def render_page_title(session) do
     session
@@ -175,6 +176,22 @@ defimpl PhoenixTest.Driver, for: PhoenixTest.Static do
     session.conn
     |> dispatch(@endpoint, method, action)
     |> maybe_redirect(session)
+  end
+
+  def open_browser(session, open_fun \\ &OpenBrowser.open_with_system_cmd/1) do
+    path = Path.join([System.tmp_dir!(), "phx-test#{System.unique_integer([:monotonic])}.html"])
+
+    html =
+      session.conn.resp_body
+      |> Floki.parse_document!()
+      |> Floki.traverse_and_update(&OpenBrowser.prefix_static_paths(&1, @endpoint))
+      |> Floki.raw_html()
+
+    File.write!(path, html)
+
+    open_fun.(path)
+
+    session
   end
 
   defp maybe_redirect(conn, session) do
--- a/test/phoenix_test/static_test.exs
+++ b/test/phoenix_test/static_test.exs
@@ -337,4 +337,35 @@ defmodule PhoenixTest.StaticTest do
                    end
     end
   end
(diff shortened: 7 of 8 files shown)


test/phoenix_test/live_test.exs
```diff
@@ -333,4 +333,24 @@ defmodule PhoenixTest.LiveTest do
       end
     end
   end
+
+  describe "preview" do
+    setup do
+      open_fun = fn path ->
+        assert content = File.read!(path)
+        assert content =~ ~r/<h1.*LiveView main page/
```
sodapopcan, line 341: I will flesh out these tests if we don't want to go with Mox.

sodapopcan:
I thought I had was that made `preview` could pretty print the `resp_body` in controller test, though it feels odd choosing something when Phoenix doesn't have anything itself.

germsvel:
Hi @sodapopcan, thanks for this PR! 🎉 

I love the idea of introducing something like this `preview` helper. And I hope to do it soon! 

But before we do that, I think I'd want to make two changes (one small and one big). 

1. The small one would be to rename it `open_browser`. As far as it makes sense, I think it's okay to keep with LiveView's convention. I like that because it'll be familiar to people and easy to guess. I like _not_ violating the principle of least surprise -- i.e. if I'm using `PhoenixTest` for the first time, and I'm acquainted with LiveView, I might reach for `open_browser` without even checking docs. But I would never guess `preview` is a helper that's availble. 
2. The larger change is to find a better implementation for the `Static` side of things. I don't want to include features that only support one side or the other, or we might get out of sync quickly and the library will lose value. So, my hope is to go into the LiveView code and see how they're doing the opening of the browser and try to replicate that. Of course, if you wanted to do that, I'd love it if someone else went and tried to figure that out. 😄

sodapopcan:
> The small one would be to rename it open_browser. 

This touches on something I brought up in another issue which is: is the overall goal of this project to be able to used in place of LiveViewTest?  I'm hoping it is 🙂 but if not, I appreciate the different naming conventions since we can just `import LiveViewTest` without conflict.  Generally I'm not a fan of blanket imports like that (I prefer to avoid `import` and `alias` whenever I can) but there are certain cases I prefer it and tests are one of them.  On a less significant note, I always like to avoid making my own helpers and `open_browser` sucks to type where as `preview` doesn't bother me, but I'm totally find with calling it `open_browser` 😄

> I don't want to include features that only support one side or the other, or we might get out of sync quickly and the library will lose value.

I very much appreciate that!

> The larger change is to find a better implementation for the Static side of things.

So I took a stab at this yesterday and got fairly close to a working version.  The main problem is the assets but I _think_ I can sort it out.  I mostly stopped as I spent a lot of time thinking about this lib yesterday to the point where I started procrastinating on the things I needed to get done.  Your response excites me though so I'm going to see if I can push it through.  I do keep wondering why such a thing isn't available in Phoenix, though 🤔

sodapopcan:
Ok, so I actually have a working prototype now haha.   I've pushed it but there some things to fix and clean up and hopefully there aren't any total blockers (again, still thinking about why Phoenix doesn't have this).  If you give me the go-ahead I'll push through and try and finish it today.

germsvel:
> Ok, so I actually have a working prototype now haha. I've pushed it but there some things to fix and clean up and hopefully there aren't any total blockers (again, still thinking about why Phoenix doesn't have this). If you give me the go-ahead I'll push through and try and finish it today.

🤩 that's awesome! Yeah, by all means, don't stop on my account. It would be great to have that helper. 

---

>  I mostly stopped as I spent a lot of time thinking about this lib yesterday to the point where I started procrastinating on the things I needed to get done

😂 sorry about that. I'm actually on the same boat. I'm having to actively _not_ spend time on the library because I'm excited about it but need to do other things too. 

---

> This touches on something I brought up in another issue which is: is the overall goal of this project to be able to used in place of LiveViewTest? 

To a large extent, the answer is yes. I'd love that. But I don't know that we'll fully be able to replace LiveView testing for two reasons: 

1. There are LiveView-specific things I don't think make sense to replace. The first that comes to mind is `assert_patch`. I don't think we should be asserting that. I just want to assert that the user sees something on the page -- whether it's done through a live patch or something else seems like an implementation detail. 
2. I envision some people wanting LiveView-specific tests as more of a "unit-test" level tests. By that I mean they may want to test specifics about how their LiveView behaves that are more LliveView-esque and less browser related. Or perhaps they need to reach out for an escape hatch, like sending messages to the LiveView process. I don't expect to replace that.
3. And then there's a gut feeling. I think LiveView might have some interactive stuff that just doesn't translate well or that we won't be able to cover perfectly _and_ have a good counterpart in the non-liveview world. I'd like to minimize this as much as possible, but there may be some things.

sodapopcan:
> 😂 sorry about that. I'm actually on the same boat. I'm having to actively not spend time on the library because I'm excited about it but need to do other things too.

😅 Not your fault!   This is one of the reasons I put off trying this out despite my initial enthusiasm on Elixir Forum.

> To a large extent, the answer is yes. I'd love that.

Ok good, me too, lol.  The immediate value I saw in this lib was the streamlined experience writing LiveView tests more so than the unification aspect (though don't get me wrong, that part is _really_ cool too).  I've been wanting something like this for a while and seen a few attempts and have made a few of my own.  But this is the first time I saw something and was like, "Ya, this is it!"  I'm a fully top-down TDD'r (I guess it's really BDD) and while LiveViewTest is super awesome, I've always seen it as more of a lower-level toolkit to enable something like this, so yay!  Thanks for really getting it going :)

-----

So back to the PR—I've got it working!

The bad news is that it involved copying over a non-insignificant amount of private functions from LiveView.  Not _too_ many, and I made sure I grokked anything I brought over.  In essence the following was needed:

- The actual function that opens the browser in a system agnostic fashion.
- The code that prefixes the `href`s and `src`s to static assets.
- Code to generate a random filename to save the page in the system's tmp dir.

On that last point, I'm not sure what the problem with using `System.unique_integer` is but I figured I'd play it safe.  I can ask on Elixir Forum.

You can find all of this code in [`PhoenixTest.Utils`](https://github.com/germsvel/phoenix_test/pull/31/files#diff-6f48bbd7da2724998ad701d6cfd997f8e792f50995a27104ceb86cea45293d08) (I'm happy to rename this or put it elsewhere).

There is the still issue that the test for the `Live` version isn't really doing anything.  We could write a better test if we brought in more LiveView but I don't think that's a good idea as it's significantly more code and I think it's better to just use `Phoenix.LiveViewTest.open_browser` directly.

sodapopcan:
Discussion around `random_id` function instead of `System.unique_integer` is found [here](https://elixirforum.com/t/question-regarding-tmp-files/61751).

lib/phoenix_test/utils.ex
```diff
@@ -0,0 +1,82 @@
...
+          if System.find_executable("cmd.exe") do
+            {"cmd.exe", ["/c", "start", path]}
+          else
+            {"xdg-open", [path]}
+          end
+      end
+
+    System.cmd(cmd, args)
+  end
+
```
sodapopcan, line 66: Based on [a convo on Elixir Forum](https://elixirforum.com/t/tmp-files-what-s-wrong-with-using-phx-system-unique-integer/61751) we could probably get away with ditching these next two functions and simply use `System.unique_integer([:monotonic])`.

germsvel: Yeah, if it seems like we could just use `System.unique_integer([:monotonic])`, I'd say let's go for it. Better to keep it simple and only introduce a more complex implementation if we need it later. 

I also went on a trip down memory lane in LiveView's git history, and it seems like that function was originally introduced for others purposes (part of the salt, I think). So, it's likely doing more than what we need. 

I do like the prefix `phx-test-` if we want to keep that. But also probably not necessary. I'll leave it up to you if you feel like it's worth including that prefix.

sodapopcan:
Ha, I just did my last force push!  I made the test for the `Live` version actually test something :)  I DI's `Phoenix.LiveViewTest.open_browser` itself and then just test that `open_browser` receives a function that is called with a `Phoenix.LiveViewTest.View`.  I'm _much_ happier with this.  In fact I'm just plain happy with it though if you know something even better, I'm all ears!

I have to leave for work soon but will address your comments later today.

Woo!

lib/phoenix_test/utils.ex
```diff
@@ -0,0 +1,82 @@
+defmodule PhoenixTest.Utils do
```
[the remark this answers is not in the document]
sodapopcan: Ya, let's just get rid of `random_id` and rename utils to `OpenBrowser`.  This is what I originally had before I brought `random_id` so we're on the same page :)

lib/phoenix_test/utils.ex
sodapopcan: I've left `phx-test-` as I believe it has value.  If for some reason someone wants to hunt down a test result in their `/tmp` then this will help them identify it.

sodapopcan:
Ok I lied, feedback has been addressed.

I hit Resolve Conversation on one comment so not sure if you'll see my answer (oh boy it's been a bit too long since I've worked on a team) but I'm leaving `phx-test-` prefix (technically the `-` comes from `System.random_integer`) since I believe it has value.  If for some reason someone wants to hunt down a test output, it'll help them identify them.  Unless you were suggesting to just call it `phx-`?  I can check now since I resolved the convo 😆

germsvel:
Thanks so much for all the great work here @sodapopcan! 🥳 

I'm going to put this in the next release. 🎉

alexslade:
@sodapopcan @germsvel Thanks for your work on this excellent library, and the work to include open_browser. 

I'm looking to add a small change, and that's to stop the removal of all `<script>` tags when creating the html file to show. Primarily because this prevents the Swoosh email previewer from looking good when opened like this; because it contains a `<script src="https://cdn.tailwindcss.com"></script>` 

I can see [where this is happening](https://github.com/germsvel/phoenix_test/blob/95e100b853ee33dda92b5279546e13d974ed11a6/lib/phoenix_test/open_browser.ex#L15), but before I dive in and issue a PR, do you have any guidance on how you would prefer this exclusion to exist? 

I was curious as to why you are excluding script files, and if you had any objections to keeping script files from external sources in place.

sodapopcan:
Hey @alexslade!

This implementation was taken from LiveView and tweaked to work with controllers.  While I did take care at the time to grok what was going on, and I specifically remember script tags being filtered out,H,H,H, that knowledge has completely vacated my brain 😅  Script tags have been filtered out since the [initial introduction](https://github.com/phoenixframework/phoenix_live_view/pull/1181/files) of the feature.  I had a poke around old issues though couldn't find anything conclusive. 
 [This is the most discussion I could find](https://github.com/phoenixframework/phoenix_live_view/issues/2330) (maybe the answer is in there?  I only skimmed it).

The best I can theorize is that since JS can't run in LiveView tests, having it run while debugging in an open browser could be potentially be very confusing.  The use-case for `open_browser` isn't meant to debug design, just to get a visual on the functionality.  CSS even often looks messed up for me in some apps depending on how I do my layouts.

I suggest bringing this up in LiveView or even just asking on the forums for a better answer than I can provide.

alexslade:
Thanks for the context ❤️ 
I'll continue to think about if / where this should be addressed.

germsvel:
Another thought off the top of my head (not sure if there's backing to it) is that allowing arbitrary JS to run could be a security concern? But... it's in tests, so not sure why that would be. Again, no backing for it. I just know that's why we'd take out script tags in other contexts -- say, when parsing markdown that allows some HTML too.

Issue #29, opened 2024-02-16
Improve documentation for function using @doc 


lcezermf:
Overall the code is pretty clean and straightforward but do you @germsvel think that would be useful to improve the doc on the modules and functions? If that is the case I can work on and submit a PR.

germsvel:
@lcezermf always happy to look at a PR that improves docs. I think those could use some love for sure. Thanks for asking!

lcezermf:
Nice, I will submit something in the next days, thanks!

germsvel:
👍 I'll go ahead and close this since it's not an active issue, but I'll be happy to review PRs!

PR #33, opened 2024-02-19
Improve doc and moduledoc

lcezermf:
This PR adds a minimal `@doc` and `@moduledoc` for the following modules:

- Query
- Assertions
- Form

Overall the other modules seem to be well-covered with documentation on the file: `lib/phoenix_test.ex`

The initial conversation was tracked on #29

### DIFF
--- a/lib/phoenix_test/assertions.ex
+++ b/lib/phoenix_test/assertions.ex
@@ -1,5 +1,9 @@
 defmodule PhoenixTest.Assertions do
-  @moduledoc false
+  @moduledoc """
+  Module for asserting HTML content in Phoenix Framework tests.
+
+  This module provides functions for asserting the presence or absence of elements with specific selectors and text content in the rendered HTML within a test session.
+  """
 
   import ExUnit.Assertions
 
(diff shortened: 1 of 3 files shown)


lib/phoenix_test/assertions.ex
```diff
@@ -1,12 +1,30 @@
 defmodule PhoenixTest.Assertions do
-  @moduledoc false
+  @moduledoc """
```
germsvel, line 2: Could we leave all `@moduledoc`s as `false`? All modules except `PhoenixTest` should be private.

lib/phoenix_test/html/form.ex
```diff
@@ -1,8 +1,31 @@
...
+    - `fields`: A list of tuples representing form fields.
+
+  ## Returns
+
+  A map representing the HTML form structure with the following keys:
+  - `"attributes"`: A map containing form attributes.
+  - `"fields"`: A list of maps representing form fields.
+  - `"operative_method"`: A string representing the form's method, extracted from hidden input or defaulting to "get".
+  """
+  @spec build(tuple()) :: map()
```
germsvel, line 28: Can we remove the `@spec`s throughout this file? We're not using dialyzer, and I don't think they provide enough value on their own.

germsvel (commented):
@lcezermf thanks for opening the PR. 

I'm sorry we got our conversation mixed. When you asked about improving documentation, I thought you meant the existing public documentation. I didn't realize you were asking about documenting private modules and functions. 

That being said, I don't mind merging this PR, but I'd like to make a couple of changes. 

1. Can we set all `@moduledoc`'s back to `false?  My intention is for those modules to remain private. They're an implementation detail and liable to change at any moment. If we make them public, and later change their API, we'd have to bump the library version but that's not really their intended use.
2. Seems like you added some `@spec`s in a module. Could we remove those? We're neither consistent in their usage (my fault), but since we're not using dialyzer, I don't find them very valuable. And I don't intend to add dialyzer in the project. Or at least, not now.

lcezermf:
Not a problem @germsvel I will fix those soon, thanks!

lcezermf:
Fixed both suggestions!

germsvel:
Thanks @lcezermf!

PR #35, opened 2024-02-20
Only document `open_browser/1`

sodapopcan:
This adds `@doc false` to `open_browser/2` (only used for tests) so that `mix docs` doesn't error out on us.

### DIFF
--- a/lib/phoenix_test.ex
+++ b/lib/phoenix_test.ex
@@ -183,6 +183,9 @@ defmodule PhoenixTest do
   |> submit_form("#user-form", name: "Aragorn")
   ```
   """
+  defdelegate open_browser(session), to: Driver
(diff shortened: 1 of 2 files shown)


sodapopcan:
@germsvel I agree about the lack of failing test.  Unfortunately I think the only way around this is to introduce some like Mox or something similar.  There is one that doesn't require a behaviour—I'd be happy to do the work if you are more comfortable with that.  I'll also think on a different dep-free way.

germsvel:
@sodapopcan thanks for the offer to do more work on this test. I think it's fine as it is. I wouldn't want to introduce Mox just for that. It was my oversight. Just the nature of having that escape hatch. Thanks for all the work!

PR #40, opened 2024-03-01
Don't raise on non-200 status code in `assert_has` and `refute_has`

### LINKED ISSUE #36: Support arbitrary status codes in `assert_has/3` and `refute_has/3`
Currently these functions assumes a status code of 200 when testing static pages, but there are some situations where we would return a non-200 code (for example authentication failure or rate limits).

I played around with it locally and was able to add an optional status argument to these functions:
https://github.com/dabaer/phoenix_test/commit/e11567597905128f2305d76c651a7670e9a0e9b6

Is this an appropriate direction to go as far as the library concerned? Or perhaps additional considerations for other functions that use `render_html` I haven't thought of?

Just wanted to get a feel before I made a pull request

Thanks!

dabaer:
Closes #36

### DIFF
--- a/lib/phoenix_test.ex
+++ b/lib/phoenix_test.ex
@@ -301,26 +301,10 @@ defmodule PhoenixTest do
   defdelegate assert_has(session, selector, text), to: Assertions
 
   @doc """
-  Performs the same assertions as above, but passes the specified status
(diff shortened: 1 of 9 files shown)


germsvel:
Thanks for this work @dabaer! 🤗

Issue #47, opened 2024-03-19
Support `phx-trigger-action`


totaltrash:
I'm not sure if this is something that can be supported but it doesn't appear to work at the moment.

I've got a form in a live view that submits to a controller when `phx-trigger-action` resolves to true, as explained [here](https://hexdocs.pm/phoenix_live_view/form-bindings.html#submitting-the-form-action-over-http). This pattern lets you use liveview for realtime form validation, then submit the form to a controller where controller stuff needs to happen (interacting with the conn). Currently the form is not submitted when triggered using phoenix_test.

It seems like something that would fit right in to the phoenix_test wheelhouse, testing an interaction between liveviews and controllers.

germsvel:
Thanks for opening this issue (and the PR) @totaltrash! I think this would be a great addition to PhoenixTest.

goncalotomas:
Just bumped into this after struggling to understand why testing out a registration form with phoenix_test wasn't working for me. I don't know exactly what's needed to add support for this, but with some guidance I would love to add a PR that fixed this.

PR #48, opened 2024-03-19
Test describing `phx-trigger-action`

totaltrash:
Here's an example describing the interaction expected for phx-trigger-action, see #47.

I'm not sure how hard implementing this would be (if possible at all)? I'm happy to have a look if you have any pointers!

### DIFF
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -126,6 +126,7 @@ defimpl PhoenixTest.Driver, for: PhoenixTest.Live do
         |> form(selector, form_data)
         |> render_submit()
         |> maybe_redirect(session)
+        |> maybe_trigger_action(selector, form_data)
 
       action_form?(form_element) ->
         session.conn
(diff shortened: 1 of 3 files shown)


germsvel:
Thanks for opening this PR @totaltrash! I think adding support for `phx-trigger-action` would be great. 

If you're up for it, I think we _can_ implement it. We already support submitting a non-LiveView form from a LiveView page, so we _should_ be able to reuse that code to do the actual form submission. 

I think the trickier part is somehow watching for forms with `phx-trigger-action`. We should do the regular `phx-submit` behavior, but when `phx-trigger-action` is set to `true`, we need to then do the non-LiveView form submission.

totaltrash:
Hi @germsvel, this is a pretty naive implementation.

It only triggers on finding the `phx-trigger-action` following a `phx-submit`, but the action could be triggered on any assigns update. It's a bit complicated because `phx-trigger-action` doesn't appear in the DOM until it's set to true (and therefore fired), so it's not something you can look for as an attribute on the form before it's triggered.

germsvel:
Got it. So, we'd have to somehow "watch" the HTML every time it's rendered and see if `phx-trigger-action` is set to `true`. That's a little more complicated, but probably doable. We' have to search the rendered HTML after _any_ type of action, clicks, submits, changes, etc. 

The behavior of `phx-trigger-action` is for the LiveView to disconnect after that and do a full form submit, right? So, once `phx-trigger-action` is set to `true`, we don't want to do anything else on the LiveView. Just submit the form and follow whatever redirect happens from it.

totaltrash:
Yes, you got it.

It may be a little more nuanced - it's a bit of an edge case but what if the updated assigns carrying the `phx-trigger-action` value has other assigns that might change values in the form - do they update before the `phx-trigger-action`?

We'll probably handle the submit as you've already described (using the submit to static form code we already have), but this is available too https://hexdocs.pm/phoenix_live_view/Phoenix.LiveViewTest.html#follow_trigger_action/2

germsvel:
Oh nice. I like [follow_trigger_action/2](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveViewTest.html#follow_trigger_action/2). I think we can definitely use that.

germsvel:
@totaltrash I'm going to close this (since this doesn't implement the feature yet) in favor of keeping the discussion in issue https://github.com/germsvel/phoenix_test/issues/47

Issue #27, opened 2024-02-15
Add an assertion helper for checking an element's children


PJUllrich:
Hey German 👋 

First of all, thank you very much for this great library! 🚀 

I have a feature request: I work a lot with lists and tables and have to assert often that e.g. the first row in a table contains a certain text. Until now, I write custom CSS selectors like e.g.:

```elixir
# Assert that the 2nd row in a table contains the string "My second row"
assert has_element?(view, ~s(table#my-table > tr:nth-of-type\(2\)), "My second row")

# Assert that the 3 row in a table has a specific data attribute
assert has_element?(view, ~s(table#my-table > tr:nth-of-type\(3\)[data-id="#{record.id}"]))
```

Do you think we could maybe add a helper that makes this easier to the library?

germsvel:
Hey @PJUllrich. Good to hear from you! Thanks for giving PhoenixTest a try an opening an issue. 

Right now, all the functions are working with CSS selectors because I find that to be very flexible because -- since it allows us to specify anything with CSS selectors. But I'd be interested to hear more about what you're looking for. 

Are you saying you'd be looking for something like one of these? Or do you have some other type of assertion in your mind? 

```elixir
|> assert_has?(within("#mytable"), "tr", "My second row")

# or

|> within("#mytable", fn session -> 
  assert_has?(session, "tr", "My second row")
end)

# or

assert_has?(child("#mytable", "tr"), "My second row")
```

PJUllrich:
hey hey, so I'm just dreaming here and am not sure whether it's feasible to implement this, but what would be great if I could "step" through the children of an element in general. So, imagine I have the following HTML:

```html
<table id="mytable">
  <tr data-id="row-1">
    <td>Hello World</td>
  </tr>
  <tr data-id="row-2">
    <td>Good bye</td>
  </tr>
</table>
```

Now, common test cases that I have to write for such HTML are:

1. Test that the first row of the table has the `data-id="row-1"`.
2. Test that the first row of the table contains the text `Hello World`

I guess, in pseudo code, this could be:

```elixir
assert_has?(nth_child(#mytable, "tr", 1), "Hello World")
```

This test would get the table and its first `tr`-child and then check whether the html of that child contains the string `Hello World`.

I don't think this could be done with CSS, but I'm unsure. Maybe Floki woudl be a better option here? That parses the HTML, so it might be easier to navigate through the children of an element.

sodapopcan:
[...]

PR #52, opened 2024-03-28
Add `at` to assert against position

germsvel:
Resolves https://github.com/germsvel/phoenix_test/issues/27

What changed?
=============

We add an `at` option to `assert_has/3` and `refute_has/3` to assert against a specific position.

This is helpful if, for example, you want to assert what is the text of the second element in a list:

```elixir
assert_has("li", at: 2, text: "Hello world")
```


sodapopcan:
@germsvel Just wanted to say I'm _super_ happy with where you landed here! 🥳

germsvel:
Thanks @sodapopcan! 🎉 Love to hear that.

PR #58, opened 2024-04-10
Fix #57: Submit pre-rendered data in static forms

### LINKED ISSUE #57: Static form submits don't include all pre-rendered data
## Expected

Given a form

```html
<form>
  <input name="input" value="value" />

  <select name="select">
    <option value="not_selected">Not selected</option>
    <option value="selected" selected>Selected</option>
  </select>

  <input name="checkbox" type="checkbox" value="not_checked" />
  <input name="checkbox" type="checkbox" value="checked" checked />

  <input name="radio" type="radio" value="not_checked" />
  <input name="radio" type="radio" value="checked" checked />
</form>
```

I expect `submit_form/3` to submit the following data

```html
input=value
select=selected
checkbox=checked
radio=checked
```

for all of these cases:
- ❌  static form in static view
- ❌  static form in live view
- ✅  live form in live view.

## Actual
For static forms (both in static and live view), only this data is submitted:

```html
radio=checked
```

ftes:


test/phoenix_test/live_test.exs
```diff
@@ -460,6 +460,17 @@ defmodule PhoenixTest.LiveTest do
       |> assert_has("#form-data", text: "email: some@example.com")
     end
 
+    test "includes pre-rendered data (input value, selected option, checked checkbox, checked radio button) via phx-submit",
```
ftes, line 463: Passes

test/phoenix_test/live_test.exs
```diff
@@ -481,6 +492,17 @@ defmodule PhoenixTest.LiveTest do
       |> assert_has("h1", text: "Main page")
     end
 
+    test "includes pre-rendered data (input value, selected option, checked checkbox, checked radio button) in regular (non phx-submit) form",
```
ftes, line 495: Fails

test/phoenix_test/static_test.exs
```diff
@@ -459,6 +459,17 @@ defmodule PhoenixTest.StaticTest do
       |> assert_has("#form-data", text: "name: Aragorn")
     end
 
+    test "includes pre-rendered data (input value, selected option, checked checkbox, checked radio button)",
```
ftes, line 462: Fails

ftes:
Absolutely, taking a look straight away.

germsvel:
Thanks! The `Form` module isn't unit tested very thoroughly (you caught me sort of in the middle of some larger refactorings that created that module). But it should mostly be tested through `live_test` and `static_test`. Let me know if you have questions, or if I can help somehow.

lib/phoenix_test/form.ex
```diff
@@ -83,18 +83,41 @@ defmodule PhoenixTest.Form do
...
       |> Html.all(selector)
       |> Enum.map(&to_form_field/1)
       |> Enum.reduce(%{}, fn value, acc -> Map.merge(acc, value) end)
 
-    Map.merge(form_data, hidden_fields)
+    Map.merge(form_data, input_fields)
+  end
+
+  defp put_form_data_select(form_data, form) do
+    # Convert to selector "select:has(option[selected][value])" once :has selector is supported by Floki
```
ftes, line 107: Would have been nicer if Floki supported `:has()` selectors.

But the way things stand we need a special `put_form_data_select` clause to cater for `select`.

lib/phoenix_test/form.ex
```diff
@@ -83,18 +83,39 @@ defmodule PhoenixTest.Form do
...
-    Map.merge(form_data, hidden_fields)
+    Map.merge(form_data, input_fields)
+  end
+
+  defp put_form_data_select(form_data, form) do
+    selects =
+      form
+      |> Html.all("select")
+      |> Enum.reduce(%{}, fn select, acc ->
+        case {Html.all(select, "option"), Html.all(select, "option[selected]")} do
```
ftes, line 111: Does not support `<select multiple>`.

lib/phoenix_test/form.ex
germsvel: Yeah, I agree. There's an open issue in Floki to suppor `has`, and I'm hoping that'll get merged someday :)

ftes:
@germsvel All done from my side.
Wdyt?

test/phoenix_test/live_test.exs
```diff
@@ -460,6 +460,18 @@ defmodule PhoenixTest.LiveTest do
...
     end
 
+    test "includes pre-rendered data (input value, selected option, checked checkbox, checked radio button) via phx-submit",
+         %{conn: conn} do
+      conn
+      |> visit("/live/index")
+      |> submit_form("#pre-rendered-data-form", [])
+      |> assert_has("#form-data", text: "input: value")
+      |> assert_has("#form-data", text: "select: selected")
+      |> assert_has("#form-data", text: "select_none_selected: first")
```
ftes, line 470: The live submit already mimicked this real-world browser behaviour (submit the first option, even if not explicitly selected).

That's how I stumbled across it.
And since that totally makes sense I replicated this behaviour for the static submits.

germsvel: Nice. Good catch! 👍

lib/phoenix_test/form.ex
germsvel: Yeah, we don't support multi selects in any of the code yet. So, I think it's fine for now _not_ to support them here. 

Just to make sure I'm understanding (for the future), is the limitation here how we transform the select data into a map where each key is the field name and the value is the option's value?

germsvel (approved):
This is excellent! 🎉  Thanks so much for making this change. 

I have a question for you in the multi-select comment for whenever you have time. But that's just to make sure I'm understanding the limitations correctly.  The PR is ready to go.

lib/phoenix_test/form.ex
ftes: > is the limitation here how we transform the select data into a map where each key is the field name and the value is the option's value

Exactly, that was my thought when I wrote that comment.

germsvel: Perfect. 👍 thanks for looping back and letting me know. I'll keep that in mind.

PR #63, opened 2024-04-13
Update assert_has examples to new API

randycoulman:
There were a few instances of example code that used the deprecated version of `assert_has/3`. This updates those to match the new API.


germsvel:
🥳 thanks for this @randycoulman!

PR #67, opened 2024-04-18
Failing test for hidden input accompanying checkbox

totaltrash:
Hi @germsvel, i noticed an issue with some of my tests when i updated phoenix test to 0.2.11.

When submitting a form with a checkbox that has an accompanying hidden input (to populate the params when the checkbox is not checked), phoenix_test seems to grab the value in the hidden input (because it's the first input in the dom?) and ignores the later checkbox regardless of whether it is checked.

### DIFF
--- a/test/phoenix_test/form_test.exs
+++ b/test/phoenix_test/form_test.exs
@@ -62,6 +62,36 @@ defmodule PhoenixTest.FormTest do
              } = form.form_data
     end
 
+    test "ignores hidden value for checkbox when checked" do
+      html = """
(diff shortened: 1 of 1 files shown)


germsvel:
Thanks for opening this @totaltrash! I'm not exactly sure how to push directly to your branch to add a fix with your tests (I think I might have to fork your repo and then push?). So instead I grabbed your tests and opened #68 to fix this. 

Thanks for the tests! 🥳 It made it so much easier to fix the issue at hand.

Issue #70, opened 2024-04-27
Select field with multiple attribute in forms raises error


sasikumar87:
Currently when encountering multi-select forms, it raises `** (ArgumentError) could not find non-disabled input, select or textarea` error

germsvel:
Hi @sasikumar87 thanks for opening this issue! I know multi-select form is something we haven't yet handled. Could you share a sample of the form select you're trying to select? It would be a helpful starting point.

sasikumar87:
I'll give an attempt to solve it first

PR #71, opened 2024-04-27
Handle multiple select in forms

sasikumar87:
Closes https://github.com/germsvel/phoenix_test/issues/70


sasikumar87:
@germsvel There are some test cases still failing. See if you have any feedbacks. I'll give another attempt tomorrow.

lib/phoenix_test/field.ex
```diff
@@ -31,8 +31,28 @@ defmodule PhoenixTest.Field do
...
+            opt = Query.find!(Html.raw(field), "option", opt)
+            Html.attribute(opt, "value")
+          end)
+
+        {true, _} ->
+          option = Query.find!(Html.raw(field), "option", option)
+          [Html.attribute(option, "value")]
+
+        {false, [_ | _]} ->
+          # TODO: raise error
```
sasikumar87, line 49: @germsvel can we raise error here or any other suggestions?

germsvel:
@sasikumar87 thanks for opening this PR! 🎉 I've given it a quick glance, but I need to give it a better look. I don't have time right now, but I'll do it as soon as I can. 

Aside from the TODO to raise an error, do you see it as done? Or are there any other open questions that I might be able to help with?

sasikumar87:
@germsvel Thanks. 
I have rebased for recent changes. With this, apart from TODO, I see it done.

lib/phoenix_test/field.ex
germsvel: @sasikumar87 if I'm following the logic correctly, then yes, we can raise an error here. This is the case when a user of the library is passing multiple options in the select but the HTML select doesn't have "multiple", right? 

Ideally, we wouldn't raise it directly in this module (since we're not raising any errors here), but it's probably fine for a first pass.

I would imagine something like: 

```elixir
msg = """
Could not find a select with a "multiple" attribute set. 

Found the following select: 

#{Html.raw(field)}
"""
raise ArgumentError, msg
```

germsvel (commented):
@sasikumar87 finally got a chance to review this PR! Thanks for all the work. 

I left a comment on the TODO section that I hope will be helpful. 

Aside from that, I noticed we don't have a test for the static-side of things. Could you add one to `static_test` and make sure your existing changes work?

lib/phoenix_test/field.ex
sasikumar87: > This is the case when a user of the library is passing multiple options in the select but the HTML select doesn't have "multiple", right?

Yes

germsvel (approved):
Love the work! 

Thanks @sasikumar87 🥳

PR #74, opened 2024-05-14
Add Wallaby driver

### LINKED ISSUE #73: Suggestion: Wallaby Driver
Would it be possible to add a Driver implementation for Wallaby?

I envision that this could be used for UI tests that involve JavaScript of any sorts, so we could test JavaScript from PhoenixTest using all the same primitives.

Once we use JavaScript in a page, we could add a test tag (e.g. `@tag :js`). For tests without the tag, it'd use the LiveView or static driver, using much less time and resources for running the test.

In that world, we could write truly user-centric tests without having to care about implementation details such as where JavaScript is used, and easily switch browser-based testing on and off.

ftes:
Fixes #73 

## Done
- add Wallaby driver
- prepare test setup for Wallaby (add app.js, convert `Endpoint` to proper/regular endpoint)
- move `Assertions` into `Driver` protocol (wrap with retry for Wallaby)
- run existing tests twice via `test_also_with_js` macro

## To do
- documentation
  - update main moduledoc: third Driver `Wallaby`, use via `@tag :js`
  - caveat: `conn` passed to `visit/2` is mostly ignored (not supported: `conn.assigns`, supported: `conn.req_cookies`)
  - possible discrepancies:
    - errors for fields without an ID (if no unique selector can be generated from other attributes) 
- fix `NoRouteError` tests (broken due to changes to `Endpoint`)
- `open_browser`
  - fill in form values (not automatically preserved in DOM tree via `value/selected/checked` attributes)
  - check stylesheets are loaded correctly
- copy initial `conn.req_cookies` via `Wallaby.Browser.set_cookie/4`


ftes:
Interesting: Some, but not all tests failing in GH actions.

```
 0 visible elements were found
```

I didn't even set up JS + chrome in `ci.yml` yet, it's a wonder any are passing 🤔 .

ftes:
Changed to `draft`, but @germsvel would love your feedback already.

arnodirlam (commented):
Nice initiative! Just a few early comments:

TODO.md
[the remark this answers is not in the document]
ftes: And implement it :)

Currently you have to pass a `%Wallaby.Session{}` instead of `%Plug.Conn{}` to `visit/2` to switch to Wallaby.

That was the easiest to do for a start, but I agree with you that a `@tag js` would be way better.
Easier to switch back and forth, more expressive.

The question then is:
Do you still always pass a `Conn` to `visit/2`? And if `@tag js: true` then that conn is simply ignored?
What about state on the `Conn` (e.g. user logged in via session cookie) - is it ok to silently ignore that too?

TODO.md
```diff
@@ -0,0 +1,14 @@
+- find inputs by exact label text (e.g. 'Race' vs. 'Race 2')
```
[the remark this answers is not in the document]
ftes: Using `PhoenixTest.Field` queries instead (operate on rendered HTML).
Benefit: Exact same behaviour for all drivers.

lib/phoenix_test/wallaby.ex
```diff
@@ -0,0 +1,259 @@
...
+  def refute_has(session, selector, opts), do: retry(fn -> Assertions.refute_has(session, selector, opts) end)
+  def assert_path(session, path), do: retry(fn -> Assertions.assert_path(session, path) end)
+  def assert_path(session, path, opts), do: retry(fn -> Assertions.assert_path(session, path, opts) end)
+  def refute_path(session, path), do: retry(fn -> Assertions.refute_path(session, path) end)
+  def refute_path(session, path, opts), do: retry(fn -> Assertions.refute_path(session, path, opts) end)
+
+  defp map_errors(fun) do
+    fun.()
+  rescue
+    e ->
```
ftes, line 237: I would prefer:
```ex
e in Wallaby.QueryError ->
  raise ArgumentError, e.message
```

But that fails to match. Why? 🤔

test/phoenix_test/live_test.exs
```diff
@@ -42,85 +46,86 @@ defmodule PhoenixTest.LiveTest do
...
 
-    test "follows push redirects (push navigate)", %{conn: conn} do
+    dual_test "follows push redirects (push navigate)", %{conn: conn} do
       conn
       |> visit("/live/redirect_on_mount/push_navigate")
       |> assert_has("h1", text: "LiveView main page")
     end
 
-    test "raises error if route doesn't exist", %{conn: conn} do
+    @tag :skip
```
ftes, line 67: Todo

test/phoenix_test/live_test.exs
```diff
@@ -398,29 +406,30 @@ defmodule PhoenixTest.LiveTest do
...
-    test "can be used to submit form", %{conn: conn} do
+    dual_test "can be used to submit form", %{conn: conn} do
       conn
       |> visit("/live/index")
       |> select("Elf", from: "Race")
       |> click_button("Save Full Form")
       |> assert_has("#form-data", text: "race: elf")
     end
 
+    # JS: multi select not supported
```
ftes, line 432: Todo

test/phoenix_test/live_test.exs
```diff
@@ -432,14 +441,15 @@ defmodule PhoenixTest.LiveTest do
...
-    test "checks a checkbox", %{conn: conn} do
+    dual_test "checks a checkbox", %{conn: conn} do
       conn
       |> visit("/live/index")
       |> check("Admin")
       |> click_button("Save Full Form")
       |> assert_has("#form-data", text: "admin: on")
     end
 
+    # JS: fix check unchecked
```
ftes, line 452: Todo

test/phoenix_test/live_test.exs
```diff
@@ -459,6 +469,7 @@ defmodule PhoenixTest.LiveTest do
   end
 
   describe "uncheck/2" do
+    # JS: fix uncheck
```
ftes, line 472: Todo

test/phoenix_test/live_test.exs
```diff
@@ -495,6 +506,7 @@ defmodule PhoenixTest.LiveTest do
   end
 
   describe "filling out full form with field functions" do
+    # JS: fix date input format
```
ftes, line 509: Todo

ftes:
Updated:
- Support `@tag :js` (via small copy&paste snippet for test module)
- Avoid duplicating tests by introducing `dual_test` macro that runs test twice: 1. with Static or Live + 2. with Wallaby driver.

ftes:
Thanks @germsvel  :)

This work is mostly done. Pretty much only documentation + `NoRouteError` fix missing from my side.

I appreciate that it is a sizeable PR which demands significant time for review though.

The only potential blocker I see is if you found any changes to existing code problematic.
E.g.
- moving `Assertions` functions to the `Driver` protocol
- changes to `Endpoint` to allow e2e tests

Apart from that, I don't see why we shouldn't move forward with this for now:
- The public API doesn't change.
- The behaviour of existing drivers doesn't change.
- Once you have the time to dig deeper, the `Wallaby` driver could always be drop-in-replaced transparently for any consumers, keeping the generic`@tag :js` in place.
  - Except of course for [Hyrum's law](https://www.hyrumslaw.com/).
  - And for support issues until then. Document as "beta" feature for now? Gather feedback to guide future initiatives?

ftes:
@germsvel Would it help you if I broke this PR into a series/stack of smaller PRs?
Wouldn't take much time on my size, and would make review less daunting. Plus might trigger some valuable discussion along the way.

germsvel:
Thanks for the offer @ftes, but don't spend your time on it just yet! I've been doing a little bit of research on playwright to see if that's a realistic alternative. 

I know you said we could introduce this and then swap it out if we need to, but once something is in the codebase, it'll be hard to remove it. Don't want to ship breaking changes too often (even in pre 1.0 version). 

Just hang in there for me if you can. I would really like to handle JS somehow -- just want to do due diligence before we fully commit to Wallaby.

ftes:
Hey @germsvel how's it going?
Still happy to split the PR. Even if you decide to go with Playwright, some preliminary work would likely be helpful: Moving Assertions into the Driver protocol.

And getting the preliminary changes in would also allow us to implement an external Wallaby Driver in the current form until playwright is usable. Without having to fork `phoenix_test`.

lib/phoenix_test/wallaby.ex
germsvel: I don't know. I would expect that to work. Is it that a different error was being raised when you tested?

germsvel:
@ftes just saw your comment after I reviewed the PR 😆 

You mentioned: 

> Still happy to split the PR. Even if you decide to go with Playwright, some preliminary work would likely be helpful: Moving Assertions into the Driver protocol.

Having reviewed the parts that I reviewed, I do think there could be some nice splits (if you're interested in doing them). Moving the assertions into the `Driver` protocol would be a good one, for sure. 

We could also add the `current_path` function as another one. 

I don't want you to do more work if you don't feel like it, but if you want to do that, that'd be a great way to introduce changes in pieces. (always easier to review too 😄 )

ftes:
Thanks for the review @germsvel

 > Having reviewed the parts that I reviewed, I do think there could be some nice splits (if you're interested in doing them). 
 
 Absolutely, I'm on it!

lib/phoenix_test/wallaby.ex
```diff
@@ -0,0 +1,266 @@
...
+      end
+
+    selector = "#{within} #{field.selector}"
+    Query.find!(html, selector)
+    Wallaby.Query.css(selector)
+  end
+
+  defp trigger_phx_change_validations(session, query) do
+    if has?(session.session, query) do
+      Wallaby.Browser.send_keys(session.session, query, [:tab])
```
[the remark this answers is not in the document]
ftes: I've been using it for a while in tests.
"Works for me", but I can't speak to how stable this is.

lib/phoenix_test/wallaby.ex
```diff
@@ -0,0 +1,266 @@
...
+      type when type in ~w(date datetime-local time week) ->
+        js = """
+          el = document.querySelector('#{field.selector}');
+          el.value = '#{value}';
+        """
+
+        session.session
+        |> Wallaby.Browser.execute_script(js)
+        |> Wallaby.Browser.find(query)
+        |> Wallaby.Element.click()
```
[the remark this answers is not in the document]
ftes: Done

lib/phoenix_test/wallaby.ex
```diff
@@ -0,0 +1,266 @@
...
+
+    session
+    |> Map.put(:last_field_query, query)
+    |> trigger_phx_change_validations(query)
+  end
+
+  def select(session, options, from: label) do
+    query = query(session, label, &Field.find_input!/2)
+
+    # Deselect other options first?
```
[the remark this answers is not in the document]
ftes: Removed the comment.

I added tests that show that Live and Static drivers both don't deselect.
Instead, multiple `select` calls result in a union of all values.

Good: This means the Wallaby driver implementation can remain simple.
Bad: There is no way to deselect options in a multi select (for any of the drivers).

test/phoenix_test/live_test.exs
ftes: Done

lib/phoenix_test/wallaby.ex
```diff
@@ -0,0 +1,266 @@
...
+  end
+
+  defp retry(fun, _now, timeout_at, interval_ms) do
+    fun.()
+  rescue
+    AssertionError ->
+      Process.sleep(interval_ms)
+      now = DateTime.to_unix(DateTime.utc_now(), :millisecond)
+      retry(fun, now, timeout_at, interval_ms)
+  end
```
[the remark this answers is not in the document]
germsvel: @ftes makes total sense. I finally had a chance to give the implementation a good look. It's really interesting how far we can get if we just render HTML with Wallaby. That being said, I'd like to explore relying on Wallaby's built-in finders as much as possible. 

I'm not asking you to do that here (though, of course, feel free to do so too), but I'm gonna create a branch based on this one and test out a few alternatives to see how that would look.

lib/phoenix_test/wallaby.ex
```diff
@@ -0,0 +1,259 @@
+defmodule PhoenixTest.Wallaby do
```
germsvel, line 1: We probably want to wrap most of this module in [Code.ensure_loaded?](https://hexdocs.pm/elixir/Code.html#ensure_loaded?/1) to check if Wallaby is included. 

We shouldn't make Wallaby a required dependency in case some people don't want to use it. What do you think?

ftes: Good point! We should definitely not force Wallaby on users.

lib/phoenix_test/wallaby.ex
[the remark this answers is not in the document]
ftes: One hard limitation you'll run into is that Wallaby doesn't provide "Exact match" finders for the input labels.
That will definitely break some tests (`Race` vs `Race 2`). Maybe worth a PR for Wallaby?

But I agree. Consistency between Drivers is nice.
But users would likely rather care more about having a feature test that simulates real user behaviour as well as possible. And using Wallaby's built-in functions where possible is probably the best way to simulate that. Instead of doing to much "HTML interpretation" ourselves.

ftes:
Did you gain any further insights by looking at Playwright, @germsvel ?

lib/phoenix_test/wallaby.ex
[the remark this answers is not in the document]
germsvel: > But users would likely rather care more about having a feature test that simulates real user behaviour as well as possible. And using Wallaby's built-in functions where possible is probably the best way to simulate that. Instead of doing to much "HTML interpretation" ourselves.

Yeah, I agree. I also imagine Wallaby's waiting mechanism logic has a few iterations on it (so it's probably ironed out any kinks),  and it's also lower level (so might be able to do things we can't).  So I'd rather rely on that instead.

ftes:
> Because I'd like to integrate both Wallaby and Playwright, I think a @tag driver: :playwright or @tag driver: :wallaby would be what I'd suggest we use.

That seems sensible.

> I also want to iron out a good API for Wallaby and Playwright. Should they start with a conn like the rest of the tests and we transform it into a session after visit/2? Or should we rely more on Wallaby's and Playwright's built-in stuff and get something like wallaby's session in the test context?

I'd say: It depends on what the goal of offering alternative drivers is.
- Is it plug-and-play interchangeability of drivers? Without the user needing to know/care what Wallaby is and how it works? Then use as little of the Wallaby/Playwright "standard" setup/assertions as possible.
- Is it to to offer the `PhoenixTest` API to users that still want maximum control of the Wallaby/Playwright internals? Then use all the "standard" stuff.

Personally, I'd vote for plug&play. For me this would be the most common requirement.
For the few use cases where I need full control (e.g. test interaction of concurrent user session) I would e.g. use vanilla Wallaby then.

> Wallaby has Wallaby.Feature that comes with the library. We don't have to use it, but I wanted to explore if we're losing anything by not using it. And does that mean we should introduce a use PhoenixTest.ConnCase or something like that? It might be helpful if we need to do some test setup.

I think this only adds support for `@sessions`. That offers the caller more flexibility (e.g. two parallel user sessions, controll session configuration).
This PR instead simply enforces a single session with default settings.

ftes:
> The playwright-elixir library has a lot of work done, but some things are still missing. The biggest issue for us is that it hasn't implemented auto-waiting assertions yet. There _are_ways to make assertions, but it's kinda clunky. So, for now, when I introduce playwright it'll be under some "experimental" section (just so people know).

That wouldn't be an issue if we use the approach in this PR: implement auto awaiting in `Driver` (wrap with retry).

config/test.exs
```diff
@@ -20,3 +20,11 @@ config :esbuild,
     cd: Path.expand("../test/assets", __DIR__),
     env: %{"NODE_PATH" => Path.expand("../deps", __DIR__)}
   ]
+
+config :wallaby,
+  otp_app: :phoenix_test,
+  js_logger: nil,
+  chromedriver: [
+    headless: System.get_env("WALLABY_HEADLESS", "t") in ["t", "true"],
```
ftes, line 28: Should go in `runtime.exs`

lib/phoenix_test/link.ex
```diff
@@ -17,7 +18,7 @@ defmodule PhoenixTest.Link do
       raw: link_html,
       parsed: link,
       id: id,
-      selector: selector,
+      selector: Element.build_selector(link),
```
ftes, line 21: Not a required change. Just for consistency with the other modules.

lib/phoenix_test/wallaby.ex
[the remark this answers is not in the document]
ftes: I wanted to give this a shot and hit a stumbling block:
CSS selectors vs XPath

- `PhoenixTest` exposes CSS selectors in the API and uses Elixir code to implement text match semantics (`exact: true/false`).
- `Wallaby` usually uses XPath to encode text match semantics

## A little background on Wallaby.Query
Wallaby mostly tries to push element searching [to the browser](https://www.selenium.dev/documentation/legacy/json_wire_protocol/#sessionsessionidelement) using `xpath` and `css` queries.
[Most of](https://github.com/elixir-wallaby/wallaby/blob/c144c92e8bc1812c1f712cc1f1c652c4d2c7ce4b/lib/wallaby/query.ex#L470-L492) the `Query` functions generate `xpath` queries.
Example: `Query.select/2` generates an xpath query with a non-exact match on the label (also supports ID filter etc.).
Nearly all `Query` functions generate non-exact xpath queries. `Query.option` is the only exception. This behaviour is not configurable. However, either wallaby could be extended to support customisable exact-match semantics or we could hand-craft the xpath queries in phoenix_test.

Wallaby supports additional query options (e.g. `text`) via Elixir post-processing.
This is less efficient, but flexible:
1. The browser executes the base query (either `css` or `xpath`).
2. The resulting element list is then filtered by inner text in Elixir.

## The dilemma
CSS selectors and text matching don't play nicely.
CSS selectors and XPath can't be combined (although there are some attempts to transpile CSS selectors to XPath, I don't know if CSS selectors are a subset of XPath though).

It's difficult to get the best of both worlds: Expose CSS selectors to PhoenixTest users and use as much vanilla Wallaby as possible. Vanilla Wallaby relies mostly on XPath.

## Where to go from here
Currently, the Wallaby driver always fetches the full HTML from the browser to do anything. That's slow.
Instead, we could at least push CSS queries to the browser where possible and filter the resulting element list via our own Elixir code (customisable exact-match semantics).
That could be a sweet spot of performance, consistency with the other drivers, and little need for changes to Wallaby itself.

I think this question might also have to be answered for Playwright.
Also, the discussion might inform what the PhoenixTest API should expose:
- generic `locator`s that can be "transpiled" to whatever works best for the driver
- CSS selectors (status quo)
 
From a quick glance, [Playwright](https://playwright.dev/python/docs/locators#locating-elements) mostly exposes generic `locator`s but does offer full control via css or xpath if need be.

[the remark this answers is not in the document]
ftes: Probably not an issue for Playwright after all. After playing with it a bit (#136) I've seen that Playwright:
- allows mixing different selector "engines" in locators (e.g. CSS + XPath)
- has in-built selectors for `label` and `text` that support exact and non-exact matching, very similar to what PhoenixTest does

ftes:
Closing in favour of #145

PR #76, opened 2024-05-14
Fix: Preserve '?' in checkbox name

ftes:
**Given**
a "static" or "live" form
```html
<input type="checkbox" name="subscribe?" value="on" />
```

**When**
I `check/2` the box and submit the form

**Then** (actual behaviour)
this is submitted
```ex
form_data = %{"subscribe" => "on"}
```

**Instead of** (expected behaviour)
```ex
form_data = %{"subscribe?" => "on"}
```


germsvel (approved):
Perfect. Thanks! 🎉

Issue #77, opened 2024-05-21
Support multiple checkboxes


totaltrash:
I'm hoping this is something that can be supported by PhoenixTest  - this strategy was supported by the removed `fill_form` but no longer works with the new form helpers (`check` etc.).

As an alternative to multiple selects, sometimes I will use multiple checkboxes to emulate the behavior of a multiple select - here's a simple example:

![image](https://github.com/germsvel/phoenix_test/assets/637350/3f5b8371-f010-4647-a0e7-f40f7e1bbd91)

```html
<div>
  <input type="checkbox" name="roles[]" id="edit_user_form_roles_admin" value="admin">
  <label for="edit_user_form_roles_admin">Administrator</label>
</div>
<div>
  <input type="checkbox" name="roles[]" id="edit_user_form_roles_user" value="user">
  <label for="edit_user_form_roles_user">User</label>
</div>
```

Phoenix/LiveView handles this just fine, converting the params provided to `phx-change` and `phx-submit` to `%{"roles" => ["admin" , ...]}`.

PhoenixTest, since the introduction of the new form handlers now results in:

```
** (ArgumentError) could not find non-disabled input, select or textarea with name "roles" within:
         <input type="checkbox" name="roles[]" id="edit_user_form_roles_admin" value="admin"/>
         <input type="checkbox" name="roles[]" id="edit_user_form_roles_user" value="user" checked="checked"/>
```

totaltrash:
Furthermore, I sometimes do similar with hidden inputs that are used to back an interactive component for searching and selecting for multiple values.

totaltrash:
I've been doing a bit of digging into how live view is handling this. Basically any param appended with `[]` is decoded into a list:

```elixir
Plug.Conn.Query.decode("roles[]=aaa&roles[]=bbb")
%{"roles" => ["aaa", "bbb"]}
```

See https://github.com/elixir-plug/plug/blob/82eb2204f59e6e2f4b40cbcab1699076f04a5758/lib/plug/conn/query.ex#L22

`Plug.Conn.Query.decode` is called from here for live view form submissions and changes:

https://github.com/phoenixframework/phoenix_live_view/blob/bc000b9d681e2456946b08e6e6ed0795bb77ff0e/lib/phoenix_live_view/channel.ex#L774

My proposal would be to treat all submitted params in this way - any param ending in `[]` should be converted to a list.

totaltrash:
[...]

PR #79, opened 2024-06-03
Failing test for multiple inputs named with []

totaltrash:
More details are in #77, but these failing tests represent the difference between how phoenix_test and phoenix handles fields named with `[]`.

### DIFF
--- a/test/phoenix_test/form_test.exs
+++ b/test/phoenix_test/form_test.exs
@@ -182,6 +182,59 @@ defmodule PhoenixTest.FormTest do
     end
   end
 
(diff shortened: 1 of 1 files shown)


germsvel:
@totaltrash thanks for adding those failing tests. Any thoughts on adding the implementation too? I'd be happy to give it a review!

ftes:
See #93 for a quick fix that satisfies the tests.

germsvel:
@totaltrash I'm going to close this since I think it's fixed by #94. Let me know if that's not the case, and we can reopen this.

Issue #72, opened 2024-04-29
support form with `?` as action


mayel:
Having a form with `action="?"` should submit to the current path but instead leads to a 404: `no route found for POST /`

mayel:
same with `action="#"`

germsvel:
Yeah, those are interesting. I think those forms of submitting forms to the current path are left over from before the HTML standard made `<form>` _not_ require an `action`. 

But I think we should support that (PhoenixTest itself requires an action, since that's probably what people want). I'd love a PR if you're up for it. Otherwise, I'll try to get that when I can.

germsvel:
@mayel is this still something you're interested in contributing with? If it's not something you need, I'll go ahead and close this issue. If it is, happy to review a PR for this.

mayel:
Yeah I'm usually happy to but don't really have the time at the moment. Also it was quite easy to workaround by changing the link. Maybe good to document at least?

germsvel:
> Also it was quite easy to workaround by changing the link

👍 sounds good. I'm going to close this then. 

> Maybe good to document at least?

I'd be happy to review a PR for docs!

PR #80, opened 2024-06-05
Fix: support links with only query parameters

jounimakela:
This commit resolves an issue where links containing only query parameters without a specified path would incorrectly direct to the root path instead of the current path, unlike standard browser behavior.

Example of a link without a path:

    <a href="?lang=en">English</a>

Previously, `click_link/2` would direct to root path. In this commit the link correctly retains the current page while applying the query parameters.

### DIFF
--- a/lib/phoenix_test/link.ex
+++ b/lib/phoenix_test/link.ex
@@ -28,4 +28,14 @@ defmodule PhoenixTest.Link do
     |> Html.attribute("data-method")
     |> Utils.present?()
   end
+
+  def maybe_append_path(link, path) do
+    link.href
+    |> URI.parse()
+    |> Map.replace_lazy(:path, fn
+      nil -> path
+      current_path -> current_path
+    end)
+    |> URI.to_string()
+  end
 end
--- a/lib/phoenix_test/static.ex
+++ b/lib/phoenix_test/static.ex
@@ -68,7 +68,9 @@ defmodule PhoenixTest.Static do
       |> dispatch(@endpoint, form.method, form.action, form.data)
       |> maybe_redirect(session)
     else
-      PhoenixTest.visit(session.conn, link.href)
+      path = Link.maybe_append_path(link, session.conn.request_path)
+
+      PhoenixTest.visit(session.conn, path)
     end
   end
 
--- a/test/phoenix_test/static_test.exs
+++ b/test/phoenix_test/static_test.exs
@@ -77,6 +77,13 @@ defmodule PhoenixTest.StaticTest do
       |> assert_has("h1", text: "LiveView main page")
     end
 
+    test "handles links with query parameters and no path", %{conn: conn} do
+      conn
+      |> visit("/page/index")
+      |> click_link("English")
+      |> assert_has("h1", text: "Main page")
+    end
+
     test "handles form submission via `data-method` & `data-to` attributes", %{conn: conn} do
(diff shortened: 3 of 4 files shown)


germsvel:
@jounimakela thanks for opening this issue. 

I think the problem isn't with `phoenix_test` itself. We're passing the `path` to [`Phoenix.ConntTest.get`](https://hexdocs.pm/phoenix/Phoenix.ConnTest.html#get/3), which seems to treat that as the root path. Would you mind opening an issue in that repo? 

I think if we want to fix that here, we should fix it for all links and not just for when we do a `click` on static pages. That means we should probably clean up the path that is passed to `PhoenixTest.visit/2`before sending it to `Phoenix.ConnTest.get/2`. 

You can see that here: https://github.com/germsvel/phoenix_test/blob/e1a240893b70448acc0146ec1830b0c1fbb789bc/lib/phoenix_test.ex#L214-L215

I'm open to the idea of us handling it here (in `visit/2`) _and_ opening an issue in Phoenix.ConnTest (just to let them know. not sure if they'd fix it). Would you be up for that?

jounimakela:
Apologies for my delayed response.

> I'm open to the idea of us handling it here (in `visit/2`) _and_ opening an issue in Phoenix.ConnTest (just to let them know. not sure if they'd fix it). Would you be up for that?

Sounds good! I'll make a proposal for a fix and open an issue.

germsvel:
Thank you! 🎉  And no apologies needed. This is open source, so we all work on it when we can. 😄 

Just wanted to check if this was still something you wanted to fix, and was doing my due diligence here.

jounimakela:
Fixing the path is now in `PhoenixTest.visit/2`. I am not too happy about having the helper function in the top-level module.

I'll also squash commits once we are happy with the changes.

Thoughts?

jounimakela:
> Were you able to open the issue on the `phoenix` repo? I'd love for this to be fixed there, and we can include the band-aid here until it's fixed. But I don't want to include the band-aid in PhoenixTest forever.

Open now! https://github.com/phoenixframework/phoenix/issues/5877

jounimakela:
@germsvel looks like `Phoenix.ConnTest.get` isn't going to change.

germsvel:
@jounimakela it's been a long time since this PR was opened. Sorry that we haven't moved it forward. 

I think that since Phoenix is not going to change how they handle this, I don't want to introduce the patch. It seems like it would be easy enough to work around by changing the link? But maybe I'm being naive about that? 

Do you have any thoughts on this? Is this still something you need? If not, I'm going to close this PR.

jounimakela:
Thanks for the follow-up.

Indeed, changing the link isn't a big deal - that's actually what we ended up doing in our code. So from a practical point of view, it's not a blocker.

That said, I still think PhoenixTest should behave like a browser as closely as possible. When test behavior diverges from how a real browser behaves, it's a false positive: the test fails but the actual application works fine.

Also, like you mentioned earlier, it's possible that a proper fix for #72 could end up solving this issue too - so maybe it would make sense to look at them together.

I'll leave it up to you how you want to proceed:

* We don't have to merge this PR as-is
* I'm happy to open a fresh issue if you'd prefer to track it that way
* I can open another PR if you have an idea for how you'd want the fix structured differently
* Or we can just forget about it - it's a minor problem with workarounds

PR #81, opened 2024-06-10
Deep merge form data in nested/prefixed forms

soundmonster:
Current implementation performs multiple passes over form defaults before merging in "user"-supplied form data.
`Map.merge` works well for flat forms where `name` attributes on inputs are not nested. When nesting is in place and
attributes look like this: `user[name]`, this approach breaks as default check boxes override hidden inputs, select
defaults override everything else, etc. 

This patch uses replaces `Map.merge` with `DeepMerge.deep_merge` for form data helpers, in the same way it's already
done in some places in the live driver.

N.B. this bug occurs for the static driver but not for the live driver that already uses `DeepMerge`.


germsvel (approved):
Love this. I keep forgetting to switch fully to DeepMerge in all of these places. Thanks for handling that!

germsvel:
@soundmonster if you have time, could you fix the conflicts? (If you don't, I'll try to get to that as soon as I can)

I think the place where there are conflicts should be somewhat easy to resolve: the multi-select should just have the `true` value, not grabbing the first one (your multi-select PR), and then everything should have DeepMerge (this PR).

PR #82, opened 2024-06-10
Do not assume first option to be selected by default for multiple select

soundmonster:
Current behavior for a multiple select is copying the default behavior for single select: assume the first option to be
selected by default. This is fine for single selects (that's how a "combo box" works after all, an "empty" options needs
to be included explicitly at the top), but this is not how multiple select works, as nothing is selected by default and
it's not safe to assume the first option to be desired.

This patch doesn't assume the first option in a multiple select to be selected by default.

### DIFF
--- a/lib/phoenix_test/form.ex
+++ b/lib/phoenix_test/form.ex
@@ -95,7 +95,7 @@ defmodule PhoenixTest.Form do
           {_, false, [only_selected]} -> Map.merge(acc, to_form_field(select, only_selected))
           {_, true, [_ | _] = all_selected} -> Map.merge(acc, to_form_field(select, all_selected))
           {[first | _], false, _} -> Map.merge(acc, to_form_field(select, first))
-          {[first | _], true, _} -> Map.merge(acc, to_form_field(select, [first]))
+          {_, true, _} -> Map.merge(acc, to_form_field(select, []))
         end
       end)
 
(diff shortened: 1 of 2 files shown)


lib/phoenix_test/form.ex
```diff
@@ -88,14 +88,14 @@ defmodule PhoenixTest.Form do
       form
       |> Html.all("select")
       |> Enum.reduce(%{}, fn select, acc ->
-        multiple = Html.attribute(select, "multiple") == "multiple"
+        multiple = !is_nil(Html.attribute(select, "multiple"))
```
soundmonster, line 91: The `multiple` attribute is binary and is considered set if present, i.e. it doesn't have to have a literal value `multiple`:
* It's set to an empty string by `phoenix_html_helpers` [here](https://github.com/phoenixframework/phoenix_html_helpers/blob/5ec48c5d0ee391429cfa080a497b490ae455ccaa/lib/phoenix_html_helpers/form.ex#L1026)
* MDN docs on the attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#multiple

germsvel: Awesome. Thanks for the links! So much easier to review that way.

germsvel (approved):
Love it! Thanks for making this change. 🎉

PR #88, opened 2024-06-17
Refactor: Move assertions into 'Driver' protocol

ftes:
Extracted from #74 

## Goal
Allow more flexibility for Driver implementations.

## Background
Browser-based testing is inherently decoupled, and usually involves some retry/await concept to ensure the rendered web UI is up to date with the server side state.

Wrapping the assertion functions in the `Driver` protocol allows flexibility for browser-based Drivers:
- use built-in framework assertions (Wallaby, Playwright) that have an integrated retry
- build custom retry wrapper around `PhoenixTest.Assertions`

### DIFF
--- a/lib/phoenix_test.ex
+++ b/lib/phoenix_test.ex
@@ -196,7 +196,6 @@ defmodule PhoenixTest do
 
   import Phoenix.ConnTest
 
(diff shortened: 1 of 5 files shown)


germsvel:
Thanks @ftes!

PR #89, opened 2024-06-17
Refactor: Add current_path to 'Driver' protocol

ftes:
Goal: Support `assert_path` implementations for browser-based `Driver` implementations

## Background
Browser-based tests have a decoupled state. It may be necessary to retry/await until the path is updated in the browser.
To support this, the `Driver` implementation should be allowed to choose how to fetch the `current_path` and not be forced to store it as a struct entry.


germsvel:
Thanks @ftes!

PR #91, opened 2024-06-17
Add regular Endpoint and esbuild

ftes:
Extracted from #74 

## Goal
Prepare for browser-based tests: Extend the test set up to allow running a Phoenix server with LiveView (javascript).


test/assets/js/app.js
```diff
@@ -0,0 +1,8 @@
+// Include phoenix_html to handle method=PUT/DELETE in forms and buttons.
+import "phoenix_html"
+import {Socket} from "phoenix"
+import {LiveSocket} from "phoenix_live_view"
+
+let csrfToken = document.querySelector("meta[name='csrf-token']").getAttribute("content")
+let liveSocket = new LiveSocket("/live", Socket, {params: {_csrf_token: csrfToken}, hooks: {SomeHook: {}, SomeOtherHook: {}}})
```
ftes, line 7: These hooks are referenced in existing tests.

test/phoenix_test/live_test.exs
```diff
@@ -61,7 +61,7 @@ defmodule PhoenixTest.LiveTest do
     end
 
     test "raises error if route doesn't exist", %{conn: conn} do
-      assert_raise Phoenix.Router.NoRouteError, fn ->
+      assert_raise ArgumentError, ~r/404/, fn ->
```
ftes, line 64: A normal endpoint catches the `NoRouteError` and tries to render it using an error view.

Since that is not implemented, we get an `ArgumentError`.

I haven't found out how to configure the `Endpoint` to emit the `NoRouteError as-is.

germsvel:
@ftes I think this might be better included as part of the Wallaby PR. Some of the changes (like setting the `server: true` in config only make sense in that context. 

I know you broke this apart so it'd be easier to review (and thank you!), but since this one is more tied to Wallaby, I think it's fine to introduce with it. I really do appreciate the other refactorings though. Those were good stuff! 👍

germsvel:
@ftes now that I've given this PR a second look, I think it'd be okay merging on its own. I like the work here, and though some stuff is closely related to trying to introduce Wallaby, I think it can go on its own. 

Do you mind trying to resolve merge conflicts and then we can get this in? 

Thanks again for all the work! 💯

ftes:
Sure, I'll resolve the merge conflicts shortly!

germsvel:
Thanks @ftes! 🎉

PR #92, opened 2024-06-17
Fix open_browser/1 example

dviramontes:
Let me know if that was the intention 🙏🏽


germsvel (approved):
That's exactly right. Thanks!

PR #93, opened 2024-06-18
Fix: Multiple inputs named with []

### LINKED ISSUE #77: Support multiple checkboxes

ftes:
Fixes #77
Replaces #79

### DIFF
--- a/lib/phoenix_test/form.ex
+++ b/lib/phoenix_test/form.ex
@@ -76,11 +76,19 @@ defmodule PhoenixTest.Form do
   end
 
   defp put_form_data(form_data, selector, form) do
+    concat_lists = fn
+      _, original, override when is_list(original) and is_list(override) ->
+        original ++ override
+
+      _, _original, _override ->
+        DeepMerge.continue_deep_merge()
+    end
+
     input_fields =
       form
       |> Html.all(selector)
       |> Enum.map(&to_form_field/1)
-      |> Enum.reduce(%{}, fn value, acc -> DeepMerge.deep_merge(acc, value) end)
+      |> Enum.reduce(%{}, fn value, acc -> DeepMerge.deep_merge(acc, value, concat_lists) end)
 
     DeepMerge.deep_merge(form_data, input_fields)
   end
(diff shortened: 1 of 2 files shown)


lib/phoenix_test/form.ex
```diff
@@ -124,6 +132,8 @@ defmodule PhoenixTest.Form do
   defp to_form_field(name_element, value_element) do
     name = Html.attribute(name_element, "name")
     value = Html.attribute(value_element, "value")
+    list? = String.ends_with?(name, "[]")
+    value = if list?, do: [value], else: value
```
ftes, line 136: @totaltrash already mentioned [Plug.Conn.Query.decode/4](https://hexdocs.pm/plug/Plug.Conn.Query.html#decode/4).

Would be great to reuse that to ensure we decode (nested) form data in the same way Phoenix and LiveView do.
Would get rid of some code here, and be more expressive.

ftes: See #94

germsvel:
@ftes I'm going to close this since we merged #94. I think this is no longer needed, right? If it is, let me know. Happy to reopen it.

PR #94, opened 2024-06-18
Build form data using Plug.Conn.Query (remove DeepMerge)

### LINKED ISSUE #77: Support multiple checkboxes

ftes:
Fixes #77
Replaces #79, #90, #93

## Goal
Ensure same form handling semantics as Phoenix controllers and LiveView.

## Details
Track form data as a flat list of `{name, value}` tuples.
This can be read straight from the DOM and needs no further conversions.
Example:
```ex
form_data = [
  {"user[name]", "Ada"},
  {"user[subscribe][]", "email"},
  {"user[subscribe][]", "push"}
]
```

Defer transforming that flat list into a nested data structure: When submitting form or evaluating other LiveLive `phx-*` callbacks.

Use [Plug.Conn.Query.decode/4](https://github.com/elixir-plug/plug/blob/v1.16.0/lib/plug/conn/query.ex#L9) to transform flat list into nested data structure. Reason: Phoenix Controllers and LiveView do the same (see https://github.com/germsvel/phoenix_test/issues/77#issuecomment-2123962071).


lib/phoenix_test/form.ex
```diff
@@ -21,6 +21,16 @@ defmodule PhoenixTest.Form do
     |> build()
   end
 
+  def build_data(%__MODULE__{} = form), do: build_data(form.form_data)
+
+  def build_data(data) when is_list(data) do
+    data
+    |> Enum.map_join("&", fn {key, value} ->
+      "#{URI.encode_www_form(key)}=#{if(value, do: URI.encode_www_form(value))}"
```
ftes, line 29: It would be nice to skip the encode/decode cycle.
But that's the only public interface`Plug.Conn.Query` seems to offer.

On the other hand, it seems reasonable to use a standardized format (`application/x-www-form-urlencoded`) as the interface.

soundmonster:
This PR does sound like a good idea to me. I [tried](https://github.com/soundmonster/phoenix_test/pull/1) supporting multiple inputs with ending with `[]` using DeepMerge, and being able to also _uncheck_ a checkbox from a group made the whole thing super complex. I tried backporting the tests from my branch on yours and they don't go through, but I guess I'll be able to get it to work on top.

lib/phoenix_test/utils.ex
```diff
@@ -1,17 +1,6 @@
...
-  def name_to_map(name, value) do
-    parts = Regex.scan(~r/[\w\?|-]+/, name)
-
-    parts
-    |> List.flatten()
-    |> Enum.reverse()
-    |> Enum.reduce(value, fn key, acc ->
-      %{key => acc}
-    end)
-  end
```
germsvel, line 13: 🔥 🔥 🔥

mix.exs
```diff
@@ -36,7 +36,6 @@ defmodule PhoenixTest.MixProject do
   defp deps do
     [
       {:ex_doc, "~> 0.31", only: :dev, runtime: false},
-      {:deep_merge, "~> 1.0"},
```
germsvel, line 39: 🎉

test/phoenix_test/static_test.exs
```diff
@@ -451,11 +451,11 @@ defmodule PhoenixTest.StaticTest do
       |> assert_has("#form-data", text: "race_2: [elf,dwarf]")
     end
 
-    test "honors empty default for multi select", %{conn: conn} do
+    test "contains no data for empty multi select", %{conn: conn} do
```
germsvel, line 454: I think the "honoring" of an empty multi select was intentionally introduced in f1461860a579d6b80468246c9a0ecee7998c59e4. Can you confirm that's no longer needed? Or will this solution equivalent?

germsvel (commented):
Thanks for opening this PR! 🥳 

I was out for a little while, so I didn't have a chance to look at it til now. 

I love this work. In my mind, this was the next logical step. we dealt with deep merging of data because of how we originally supported map data being passed for forms with `fill_form`. 

But since we've moved to using labels, it makes sense we keep the names flat as much as possible and encode/decode that stuff only when we need to send it to the server. So, I'm glad to see this coming through! 👍 

I had one question about a test that was modified (see the comment on that question), and I want to play around locally with this branch a little more before merging. But I think this is almost ready to go. 

Thanks again!

test/phoenix_test/static_test.exs
soundmonster: Yes can confirm that clearing multi-select values only works if an explicitly empty value is hidden next to it, just like with checkboxes 👍

ftes: So you agree it is correct to change the test? @soundmonster

soundmonster: Yes, absolutely!

ftes: Then let's get this merged @germsvel , wdyt?

germsvel: Nice! 🎉  Thanks for checking on this.

PR #100, opened 2024-07-19
Allow non-string data types as inputs to fields.

### LINKED ISSUE #98: Should non-text fields support corresponding types?
For input fields other than "text", should PhoenixTest support the corresponding data type?

```elixir
session
# this works
|> fill_in("Number input", with: "10")

# this doesn't
|> fill_in("Number input", with: 10)
# ** (FunctionClauseError) no function clause matching in URI.encode_www_form/1
#
#     The following arguments were given to URI.encode_www_form/1:
#
#         # 1
#         10
#
#     Attempted function clauses (showing 1 out of 1):
#
#         def encode_www_form(string) when is_binary(string)

# analogously, this works
|> fill_in("Date input", "2023-12-30")
# while these don't
|> fill_in("Date input", Date.utc_today())
|> fill_in("Date input", Date.new!(2023, 12, 30))
```

A possible solution would be to add a `type` field to the `Field` struct and then have a check with possible conversion in `Field.to_form_data/1`.

I can draft a fix, the question is:
* is this a thing PhoenixTest should support?
* what other input types will need special handling, besides `number` and `date`?
* is this a bottomless pit of maintenance pain, or a nice feature that enables more fluent tests?

soundmonster:
Fixes #98

### DIFF
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -102,7 +102,7 @@ defmodule PhoenixTest.Live do
     session
     |> render_html()
     |> Field.find_input!(label)
-    |> Map.put(:value, value)
+    |> Map.put(:value, to_string(value))
     |> then(&fill_in_field_data(session, &1))
   end
 
--- a/lib/phoenix_test/static.ex
+++ b/lib/phoenix_test/static.ex
@@ -118,7 +118,7 @@ defmodule PhoenixTest.Static do
     session
     |> render_html()
     |> Field.find_input!(label)
-    |> Map.put(:value, value)
+    |> Map.put(:value, to_string(value))
     |> then(&fill_in_field_data(session, &1))
   end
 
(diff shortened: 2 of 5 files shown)


soundmonster:
PR number 💯 🥳

germsvel (commented):
Thanks for making these changes. 

I wonder if we can make this simpler by changing `fill_in` in https://github.com/germsvel/phoenix_test/blob/b79e8a3957506f52c7c2a104145a0a639854e614/lib/phoenix_test/live.ex#L105  (and the static equivalent) to call `to_string/1` instead of doing it in `build_data`: 

```elixir
  def fill_in(session, label, with: value) do
    session
    |> render_html()
    |> Field.find_input!(label)
    |> Map.put(:value, to_string(value))
    |> then(&fill_in_field_data(session, &1))
  end
```

soundmonster:
I considered that but it felt safer to change one place instead of two. But I can see how keeping this kind of casting logic at the perimeter makes sense, keeping the internal data structure string-only. I'll try that, thank you!

soundmonster:
@germsvel the conversion to strings is happening in the `fill_in` function now, in the two adapters, please have another look.

germsvel (approved):
Love it! Thanks for making the change. 🥇

PR #102, opened 2024-08-01
Issues with unchecking checkboxes and unselecting items from multiple select

totaltrash:
This commit has some failing tests for a number of issues, but they all relate to removing items from multiple selects or checkbox groups.

1. Can't unselect options from a multiple select (maybe we need an `unselect()` helper)
2. Can't uncheck checkboxes that aren't backed by a hidden input
3. A form prefilled with one selected item in a multiple select has duplicate items (same for checkbox group)

I've had a quick go at an implementation to fix these issues, but I can't quite get my head around what you would need to do to remove items from a form in a session. I'll give it another crack soon, but any pointers would be awesome

### DIFF
--- a/test/phoenix_test/live_test.exs
+++ b/test/phoenix_test/live_test.exs
@@ -444,10 +444,24 @@ defmodule PhoenixTest.LiveTest do
     test "works for multiple select", %{conn: conn} do
       conn
       |> visit("/live/index")
-      |> select("Elf", from: "Race")
       |> select(["Elf", "Dwarf"], from: "Race 2")
       |> click_button("Save Full Form")
-      |> assert_has("#form-data", text: "[elf, dwarf]")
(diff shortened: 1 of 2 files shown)


test/phoenix_test/live_test.exs
```diff
@@ -444,10 +444,24 @@ defmodule PhoenixTest.LiveTest do
...
+      |> click_button("Save Full Form")
+      |> assert_has("#form-data", text: "race_2: [elf]")
+    end
+
+    test "deselect all from multiple select", %{conn: conn} do
+      conn
+      |> visit("/live/index")
+      |> select([], from: "Race 2")
+      |> click_button("Save Full Form")
+      |> assert_has("#form-data", text: "race_2: []")
```
totaltrash, line 464: I _think_ this is what you would expect from submitting an empty multiple select?

PR #108, opened 2024-08-28
Add upload/2 to test file uploads

ftes:
Add `PhoenixTest.upload/2` to test live and static file uploads.


lib/phoenix_test/live.ex
```diff
@@ -101,6 +101,35 @@ defmodule PhoenixTest.Live do
...
 
+  def upload(session, field) do
+    path = field.value
+    file_stat = File.stat!(path)
+    file_name = Path.basename(path)
+    form = Field.parent_form!(field)
+    live_upload_name = String.to_existing_atom(field.name)
+
+    mime_type =
+      case path |> Path.extname() |> String.downcase() do
```
ftes, line 112: Use `mime` as optional dependency to improve matching?
Or require caller to provide metadata (though for a feature test I'd like to use the actual file).

https://hex.pm/packages/mime

lib/phoenix_test/live.ex
```diff
@@ -101,6 +101,35 @@ defmodule PhoenixTest.Live do
     fill_form(session, form.selector, form_data, additional_data)
   end
 
+  def upload(session, field) do
+    path = field.value
+    file_stat = File.stat!(path)
+    file_name = Path.basename(path)
+    form = Field.parent_form!(field)
+    live_upload_name = String.to_existing_atom(field.name)
```
ftes, line 109: Works for the simple only test so far.

Will it work in other scenarios (e.g. `name="user[avatar]"`)? Probably not.

lib/phoenix_test/live.ex
ftes: Also, extract to module (duplicated in `Static`)

lib/phoenix_test/static.ex
```diff
@@ -116,6 +116,20 @@ defmodule PhoenixTest.Static do
...
+    path = field.value
+
+    mime_type =
+      case path |> Path.extname() |> String.downcase() do
+        ".jpg" -> "image/jpeg"
+        ".jpeg" -> "image/jpeg"
+      end
+
+    upload = %Plug.Upload{content_type: mime_type, filename: Path.basename(path), path: path}
+    field = %{field | value: upload}
```
ftes, line 129: We're piggybacking on regular `Form.form_data` and storing a non-string value.
Later in `submit/submit_form` we have to pull "regular values" and "uploads" apart again.

lib/phoenix_test/static.ex
```diff
@@ -173,16 +187,49 @@ defmodule PhoenixTest.Static do
...
 
+  defp build_data(form_data) do
+    {uploads, regular} = Enum.split_with(form_data, fn {_, value} -> is_struct(value, Plug.Upload) end)
+
+    regular
+    |> Form.build_data()
+    |> inject_uploads(uploads)
+  end
+
+  defp inject_uploads(map, uploads) do
```
ftes, line 208: Roundabout way of converting `user[name]` to `["user", "name"]` using `Plug.Conn.Query.decode` and a magic `"placeholder"` value.

ftes: If we go down this route, this should be extracted to its own module.

Alternatives:
- custom parsing of `name` to `path` (e.g. `user[name]` to `["user", "name"]`)
- serialize & deserialize a multipart request (closer to real world, but too much effort?!)

lib/phoenix_test/static.ex
ftes: Alternatives
- Store in new `FormData.uploads` field to store this separately
- Store in new `Static.uploads` field (but have to keep track of which form they belong to)

lib/phoenix_test/live.ex
ftes: Extracted and added optional dep.

lib/phoenix_test/live.ex
ftes: I think we're fine.

mix.exs
```diff
@@ -48,7 +48,8 @@ defmodule PhoenixTest.MixProject do
       {:phoenix_live_view, "~> 0.20.1"},
       {:esbuild, "~> 0.8", only: :test, runtime: false},
       {:styler, "~> 0.11", only: [:dev, :test], runtime: false},
-      {:plug_cowboy, "~> 2.7", only: :test, runtime: false}
+      {:plug_cowboy, "~> 2.7", only: :test, runtime: false},
+      {:mime, ">= 1.0.0", optional: true}
```
ftes, line 52: `plug` depends on `mime` anyway

https://github.com/elixir-plug/plug/blob/main/mix.exs#L47

lib/phoenix_test/live.ex
```diff
@@ -101,6 +102,30 @@ defmodule PhoenixTest.Live do
...
+      size: file_stat.size,
+      type: mime_type
+    }
+
+    session.view
+    |> file_input(form.selector, live_upload_name, [entry])
+    |> render_upload(file_name)
+    |> maybe_redirect(session)
+
+    session
```
germsvel, line 126: Should `maybe_redirect` be the last step in the pipeline? That's typically how we can follow the redirect.

lib/phoenix_test/static.ex
germsvel: I think I like the first alternative -- keeping it in a new `FormData.uploads` field. I think that'll make processing the upload clearer. No need to overload value.

test/test_helper.exs
```diff
@@ -1,3 +1,4 @@
 ExUnit.start()
 
+{:ok, _} = Supervisor.start_link([{Phoenix.PubSub, name: PhoenixTest.PubSub}], strategy: :one_for_one)
```
germsvel, line 3: Why do we need to start PubSub here?

lib/phoenix_test/static.ex
germsvel: Do you think we could move this `inject_uploads` into `Form`? I know this is only used for "Static" uploads, but I think all the `build_data` for the form should live there. 

And then, could we add a test for this `inject_uploads` directly? I think that would clarify that the reason for "placeholder" (or at least, it'll make it clear that the "placeholder" gets dropped when encoding the paths).

test/support/page_view.ex
```diff
@@ -216,6 +216,21 @@ defmodule PhoenixTest.PageView do
...
+      id="file-upload-form"
+      method="post"
+      action="/page/create_record"
+      enctype="multipart/form-data"
+    >
+      <label for="avatar">Avatar</label>
+      <input id="avatar" name="avatar" type="file" />
+
+      <label for="nested_avatar">Nested Avatar</label>
+      <input id="nested_avatar" name="user[avatar]" type="file" />
```
germsvel, line 229: 👍 I see you tested the nested form here. Nice.

lib/phoenix_test.ex
```diff
@@ -627,6 +627,38 @@ defmodule PhoenixTest do
...
+  ```
+
+  We can upload a file:
+
+  ```elixir
+  session
+  |> upload("Avatar", "/path/to/file")
+  ```
+  """
+  def upload(session, label, path) do
```
germsvel, line 654: 🎉  love the API for `upload/3` here.

germsvel (commented):
Love the work! 🥳 

Thanks for opening this PR. I've had uploads in the back of my head for a little while now. 

I left a few minor comments. Let me know if anything is unclear.

lib/phoenix_test/static.ex
ftes: Both good ideas 👍

test/test_helper.exs
ftes: Otherwise the live upload test fails.

```ex
** (ArgumentError) unknown registry: PhoenixTest.PubSub
    (elixir 1.16.0) lib/registry.ex:1391: Registry.info!/1
    (elixir 1.16.0) lib/registry.ex:1007: Registry.register/3
    (phoenix_pubsub 2.1.3) lib/phoenix/pubsub.ex:121: Phoenix.PubSub.subscribe/3
    (phoenix 1.7.10) lib/phoenix/test/channel_test.ex:406: Phoenix.ChannelTest.subscribe_and_join/4
    (phoenix_live_view 0.20.3) lib/phoenix_live_view/test/upload_client.ex:137: Phoenix.LiveViewTest.UploadClient.build_and_join_entry/3
    (phoenix_live_view 0.20.3) lib/phoenix_live_view/test/upload_client.ex:68: anonymous fn/5 in Phoenix.LiveViewTest.UploadClient.handle_call/3
    (elixir 1.16.0) lib/enum.ex:2528: Enum."-reduce/3-lists^foldl/2-0-"/3
    (phoenix_live_view 0.20.3) lib/phoenix_live_view/test/upload_client.ex:64: Phoenix.LiveViewTest.UploadClient.handle_call/3
```

https://github.com/phoenixframework/phoenix_live_view/blob/v0.20.17/lib/phoenix_live_view/test/upload_client.ex#L139

I didn't check whether there's a way to avoid this.

lib/phoenix_test/static.ex
ftes: Moved to Form and added tests

test/phoenix_test/form_test.exs
```diff
@@ -341,4 +341,42 @@ defmodule PhoenixTest.FormTest do
...
+    test "injects nested upload" do
+      form_data = %{"user" => %{"name" => "Frodo"}}
+      uploads = [{"user[avatar]", %Plug.Upload{}}]
+
+      injected = Form.inject_uploads(form_data, uploads)
+
+      assert injected == %{"user" => %{"name" => "Frodo", "avatar" => %Plug.Upload{}}}
+    end
+
+    test "only handles first file in list" do
```
ftes, line 373: Todo: Handle multiple files in list.

test/phoenix_test/form_test.exs
```diff
@@ -341,4 +341,54 @@ defmodule PhoenixTest.FormTest do
...
+    test "handles all files in list" do
+      form_data = %{}
+      uploads = [{"avatar[]", upload(0)}, {"avatar[]", upload(1)}]
+
+      injected = Form.inject_uploads(form_data, uploads)
+
+      assert injected == %{"avatar" => %{0 => upload(0), 1 => upload(1)}}
+    end
+
+    @tag :skip
```
ftes, line 382: image

I'm unsure if not starting at `0` would ever trip up a real-world changeset.

test/phoenix_test/static_test.exs
```diff
@@ -540,6 +540,34 @@ defmodule PhoenixTest.StaticTest do
...
+  describe "upload/2" do
+    test "uploads image", %{conn: conn} do
+      conn
+      |> visit("/page/index")
+      |> upload("Avatar", "test/files/elixir.jpg")
+      |> click_button("Save File upload Form")
+      |> assert_has("#form-data", text: "avatar: elixir.jpg")
+    end
+
+    test "uploads image list", %{conn: conn} do
```
ftes, line 552: Added support for lists `avatars[]`.

lib/phoenix_test/live.ex
ftes: Done

lib/phoenix_test/static.ex
ftes: Done in 7ce41310

ftes:
Rebased and addressed feedback. Good for another review I reckon @germsvel

test/test_helper.exs
germsvel: Yeah, seems like `LiveViewTest` uses `PubSub` (probably for upload progress). 

My question now is, should we be starting `PubSub` in the test app instead of `test_helper.exs`? Put differently, I don't want people who use `phoenix_test` to have to start` PubSub` in their test helper. I imagine they _wouldn't_ have to since Phoenix apps already have it started. But it would be nice for our `test_helper.exs` to be more or less like their helper would be. 

Does that make sense?

lib/phoenix_test/static.ex
```diff
@@ -161,6 +162,21 @@ defmodule PhoenixTest.Static do
...
+    mime_type = FileUpload.mime_type(path)
+    upload = %Plug.Upload{content_type: mime_type, filename: Path.basename(path), path: path}
+    field = session |> render_html() |> Field.find_input!(label)
+    form = Field.parent_form!(field)
+
+    Map.update!(session, :active_form, fn active_form ->
+      if active_form.selector == form.selector do
+        ActiveForm.add_upload(active_form, {field.name, upload})
+      else
+        ActiveForm.new(id: form.id, selector: form.selector, uploads: [{field.name, upload}])
```
germsvel, line 175: I think in this case we'd want to make sure we're adding any form data that is defaulted on the page -- like a pre-selected checkbox, for example. 

I think we can do it like this? 

```elixir
ActiveForm.new(id: form.id, selector: form.selector)
|> ActiveForm.prepend_form_data(form.form_data) # <- add default data first
|> ActiveForm.add_upload([{field.name, upload}] 
```

What do you think? Are we missing anything else when creating a new active form?

germsvel (commented):
Only had a chance to give it a quick review right now, but wanted to share what I found anyway. 

Mostly a question about where we should put the `PubSub` and a note on making sure we're prefilling all the default values in a form.

test/test_helper.exs
ftes: > I imagine they wouldn't have to since Phoenix apps already have it started

I also think any user will already have it started.

If you want `test_helper.exs` to serve as a blueprint for users of this library, then I get what you're saying.
So far, there's nothing interesting in there that anybody would want to look at.

lib/phoenix_test/static.ex
ftes: Good point. Maybe there should be a shared helper function called both by `fill_field_data` and `upload`.

ftes: Extracted `new_active_form`.

ftes: And now moved to `ActiveForm.new/1`

test/test_helper.exs
germsvel: > If you want test_helper.exs to serve as a blueprint for users of this library, then I get what you're saying.

No, it's not about having `test_helpers.exs` be a blueprint for users of the library. 

It's more about making sure that `phoenix_test` tests run the same as they would in a project that includes `phoenix_test`. That way, we can be confident that shipping the library won't behave differently in projects from what our tests have. Make sense?

lib/phoenix_test/active_form.ex
```diff
@@ -1,19 +1,29 @@
...
-  def new(opts \\ []) do
-    Map.merge(%{id: nil, selector: nil, form_data: []}, Map.new(opts))
+  alias PhoenixTest.Form
+
+  defstruct [:id, :selector, form_data: [], uploads: []]
+
+  def new(form_or_opts \\ [])
+
+  def new(%Form{} = form) do
+    %__MODULE__{id: form.id, selector: form.selector, form_data: form.form_data}
```
germsvel, line 11: 👍 I like this! I see how it cleaned up the `prepend_data` calls.

germsvel (commented):
A few last comments. I'm excited for this! 

Thanks again for putting this together! 🏅

test/phoenix_test/form_test.exs
ftes: I just realised I made a mistake.
I was thinking about what input `name`s phoenix generates in [inputs_for](https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#inputs_for/1).

But that need not concern us here. We should just take the names as is:
- HTML `name="user[avatars][0][file]` => params `%{"user" => %{"avatars" => %{"0" => %{"file"=> %Plug.Upload{}`
  - I think this is what `inputs_for` does by default, and we should support it
  - Ecto changesets then gracefully handle this nested `%{string_index => %{` (I think, didn't double check now)
- HTML `name="users[avatars][]` => params `%{"user" => %{"avatars" => [%Plug.Upload{}`

test/test_helper.exs
ftes: Makes sense.
In that case, we should have a `PhoenixTest.TestApplication`, started only during tests. With children:
- `Endpoint`
- `PubSub`

test/phoenix_test/form_test.exs
ftes: This is what I meant: https://github.com/ftes/phoenix_live_view/blob/83f3f84096423129d4d7e3f25b676997884d7e35/test/phoenix_component/components_test.exs#L515-L520

ftes: Updated PR to support both. Added tests for both.

Can't use `put_in` anymore, which makes the recursive helper function more cryptic. The tests should be explanatory though.

test/test_helper.exs
ftes: I haven't been able to find a way of starting a test-only application. Maybe you have an idea @germsvel ?

germsvel: No, I think you were right the first time. I'm not sure why, but I _thought_ we were starting the app somewhere already (including the endpoint). But I just gave this another look and see that we're doing that here. 

Given that we've already started here, I think it's fine to move forward with this now. If I have time later, I might try to explore moving these to a test-only application (or if that's even desired)

germsvel (approved):
This looks ready to go! Thanks again for all the work! 🥇

PR #109, opened 2024-08-30
Handle checkbox phx-click outside of form

### LINKED ISSUE #101: Handle checkbox with phx-click but no form
Checkboxes (and other inputs) may be used outside of forms to trigger `phx-click` and `phx-change` events that are specified on the input (not on the form).

Either
1. `check/2` could support this use case or (preferred)
2. add a dedicated `click_checkbox/2`

In both cases, the `name` attribute should not be required.

## Current behaviour
Calling `check/2` on a checkbox outside of a form results in

```ex
     ** (ArgumentError) Could not find "form" for an element with selector "#my-checkbox-id".

     Found other potential "form":
```

If there is no `name` attribute on that checkbox, it fails even earlier:

```ex
     ** (FunctionClauseError) no function clause matching in Regex.scan/3
```

ftes:
Fix #101

## Out of scope
While this fixes this specific case, a more thorough investigation and solution might be better:
- `phx-change` on `<input type="text"` without a form
- `phx-click` on a checkbox inside a form (should trigger both events: `phx-click` on checkbox AND `phx-change` on form)

### DIFF
--- a/lib/phoenix_test/form.ex
+++ b/lib/phoenix_test/form.ex
@@ -15,6 +15,13 @@ defmodule PhoenixTest.Form do
     |> build()
   end
 
+  def find_by_descendant(html, descendant) do
+    case Query.find_ancestor(html, "form", descendant_selector(descendant)) do
+      {:found, element} -> {:ok, build(element)}
+      other -> {:error, other}
+    end
+  end
+
   def find_by_descendant!(html, descendant) do
     html
     |> Query.find_ancestor!("form", descendant_selector(descendant))
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -87,18 +87,30 @@ defmodule PhoenixTest.Live do
     active_form = session.active_form
     existing_data = active_form.form_data
     new_form_data = Field.to_form_data(field)
-    form = Field.parent_form!(field)
 
-    form_data =
-      if active_form.selector == form.selector do
-        existing_data ++ new_form_data
-      else
-        new_form_data
-      end
+    case Field.parent_form(field) do
+      {:ok, form} ->
+        form_data =
+          if active_form.selector == form.selector do
+            existing_data ++ new_form_data
+          else
+            new_form_data
+          end
 
-    additional_data = %{"_target" => field.name}
+        additional_data = %{"_target" => field.name}
 
-    fill_form(session, form.selector, form_data, additional_data)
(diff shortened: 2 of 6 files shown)


lib/phoenix_test/form.ex
```diff
@@ -15,6 +15,13 @@ defmodule PhoenixTest.Form do
     |> build()
   end
 
+  def find_by_descendant(html, descendant) do
```
germsvel, line 18: Could we add a test for this?

lib/phoenix_test/live.ex
```diff
@@ -87,18 +87,30 @@ defmodule PhoenixTest.Live do
...
+            new_form_data
+          end
 
-    additional_data = %{"_target" => field.name}
+        additional_data = %{"_target" => field.name}
 
-    fill_form(session, form.selector, form_data, additional_data)
+        fill_form(session, form.selector, form_data, additional_data)
+
+      _no_form ->
```
germsvel, line 104: What do you think about matching here on `{:error, _error}` to make it more explicit? I'm not even sure what the `_error` would be in that case. If we know, we could make it clear -- `{:error, :no_form}` (or something like that.

germsvel (commented):
Love it! 👍 

Left a few comments, but they're all small things (I hope). Thanks for the work on this! 

(I'll add an issue to try to keep track of all the other fields that should work outside of forms with `phx-change`, etc.)

germsvel:
While I was looking at implementing other inputs (like radio buttons), I realized that we have some prior art in this. When we click a button, we have to determine if it's a `phx-click` or inside a form. See https://github.com/germsvel/phoenix_test/blob/57f9a488c33172a8fd7626ddf144e37488b091b0/lib/phoenix_test/live.ex#L58-L83

I like two things about that approach: 

1. I like how there's a `cond` statement where the final step is a clear instruction to the user. It gives a nice error of what we're expecting. 
2. It has clear path separation at the top level (in the `click_button` function). I made it difficult for us do that in `fill_in` when I extracted common code into the `PhoenixTest` module. So, I should probably revert that, since that the `Live` and `Static` implementations will differ now.

germsvel:
@ftes I'm going to close this in favor of https://github.com/germsvel/phoenix_test/pull/116. I ended up handling `check` and `uncheck` there. It lines up with how I've handled radio buttons and select dropdowns. 

Thank you for opening this and for all the hard work!

ftes:
#116 better, thanks for letting me know 👍

Issue #105, opened 2024-08-06
Error when hidden field is updated by server


totaltrash:
I'm getting an error when updating a second field in a nested form:

```elixir
session
# This works
|> fill_in("First Name", with: "Some")
# This fails:
|> fill_in("Last Name", with: "Contact")
```

```
** (ArgumentError) value for hidden "form[involved][0][_touched]" must be one of ["_form_type,_persistent_id,_touched,contact,involved_how"], got: "_form_type,_touched,contact"
code: |> fill_in("Last Name", with: "Contact")
     stacktrace:
       (phoenix_live_view 0.20.17) lib/phoenix_live_view/test/live_view_test.ex:1102: Phoenix.LiveViewTest.call/2
       (phoenix_test 0.3.1) lib/phoenix_test/live.ex:172: PhoenixTest.Live.fill_form/4
       test/features/enquiry_test.exs:32: (test)
```

totaltrash:
I've been unable to replicate this with a test unfortunately. I'll have another crack later this week hopefully

totaltrash:
I've reproduced the issue here: https://github.com/totaltrash/phoenix_test_playground.

The issue isn't about nested forms, it's about hidden fields that are updated during a `phx-change` event. The pertinent bits are:

https://github.com/totaltrash/phoenix_test_playground/blob/master/test/my_app_web/features/smoke_test.exs

```elixir
  test "admin can create user", %{conn: conn} do
    conn
    |> visit("/users/new")
    |> fill_in("Name", with: "Aragorn") # <- this works fine
    |> fill_in("Email", with: "aragorn@gondormail.com") # <- this fails
    |> click_button("Save")
    |> assert_has("table", text: "Aragorn")
  end
```

The second form change fails with:

```
** (ArgumentError) value for hidden "user[secret]" must be one of ["z"], got: ""
code: |> fill_in("Email", with: "aragorn@gondormail.com")
     stacktrace:
       (phoenix_live_view 1.0.0-rc.6) lib/phoenix_live_view/test/live_view_test.ex:1102: Phoenix.LiveViewTest.call/2
       (phoenix_test 0.3.1) lib/phoenix_test/live.ex:121: PhoenixTest.Live.fill_form/4
       test/my_app_web/features/smoke_test.exs:8: (test)
```

Which appears to be LiveViewTest complaining about changing the value of a hidden as if it was a normal text field (not through the `hidden_fields` in `render_change`)

The hidden field is changed here:

[...]

PR #110, opened 2024-09-01
Fix multi checkbox: Add hidden input only once

ftes:
## Given
```html
<input type="hidden" name="race_3[]" value="dwarf" />
<label><input type="checkbox" name="race_3[]" value="human" />Human</label>
<label><input type="checkbox" name="race_3[]" value="elf" />Elf</label>
```

```ex
|> check("Human")
|> check("Elf")
```

## Expected
```ex
race_3: [dwarf, human, elf]
```

## Actual
```ex
race_3: [dwarf, dwarf, dwarf, human, elf]
```

## Todo
- [x] Add failing test
- [ ] Fix

### DIFF
--- a/lib/phoenix_test.ex
+++ b/lib/phoenix_test.ex
@@ -627,6 +627,7 @@ defmodule PhoenixTest do
     |> Field.find_input!(label)
     |> then(&Driver.fill_in_field_data(session, &1))
   end
+
   @doc """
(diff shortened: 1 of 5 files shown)


germsvel:
Not 100% sure, but I think this might be related to #105. Might not be 100% the same, but I think when we moved active form data from being maps into tuples, we lost the ability to dedup. 

In any case, I'm looking into a solution for #105 that might fix this too.

PR #111, opened 2024-09-01
Add assert_download/2 to test file downloads

ftes:
Assert helper to verify a file download.

  > ### Note on supported download types
  >
  > Only downloads with HTTP response header `Content-Type: attachment; filename=...` are supported.

  ## Examples

  ```elixir
  conn
  |> visit("/users")
  |> click_link("Download avatar")
  |> assert_download("avatar.jpg")

  conn
  |> visit("/users")
  |> click_link("Download avatar")
  |> assert_download(fn %{mime_type: type, content: content} ->
    assert file.mime_type == "image/jpeg"
    assert content == File.read!(expected_file_path)
  end)
  ```

### DIFF
--- a/lib/phoenix_test/assertions.ex
+++ b/lib/phoenix_test/assertions.ex
@@ -188,6 +188,23 @@ defmodule PhoenixTest.Assertions do
     session
   end
 
+  def assert_download(session, _) when not is_struct(session, PhoenixTest.Static) do
+    raise ArgumentError, "Only downloads via Phoenix.Controller are supported."
+  end
+
+  def assert_download(session, fun) when is_function(fun, 1) do
+    session |> PhoenixTest.Static.download_file() |> fun.()
+
+    session
+  end
+
+  def assert_download(session, file_name) when is_binary(file_name) do
+    file = PhoenixTest.Static.download_file(session)
+    assert file.name == file_name
+
+    session
+  end
+
   def assert_path(session, path) do
     uri = URI.parse(PhoenixTest.Driver.current_path(session))
 
--- a/lib/phoenix_test/static.ex
+++ b/lib/phoenix_test/static.ex
@@ -3,6 +3,7 @@ defmodule PhoenixTest.Static do
 
   import Phoenix.ConnTest
 
+  alias ExUnit.AssertionError
   alias PhoenixTest.ActiveForm
   alias PhoenixTest.DataAttributeForm
   alias PhoenixTest.Element.Button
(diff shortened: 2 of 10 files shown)


test/phoenix_test/static_test.exs
```diff
@@ -736,4 +738,29 @@ defmodule PhoenixTest.StaticTest do
...
+      conn
+      |> visit("/page/download")
+      |> assert_download("elixir.jpg")
+    end
+
+    test "custom assertion via function", %{conn: conn} do
+      conn
+      |> visit("/page/download")
+      |> assert_download(fn file ->
+        assert file.mime_type == "image/jpeg"
```
ftes, line 753: We could add a third options argument to `assert_download` that accepts `mime_type` as an option.

lib/phoenix_test/static.ex
```diff
@@ -50,6 +52,20 @@ defmodule PhoenixTest.Static do
     end
   end
 
+  def download_file(%{conn: conn}) do
+    file_name =
+      case Plug.Conn.get_resp_header(conn, "content-disposition") do
+        ["attachment; filename=\"" <> name | _] -> String.trim_trailing(name, "\"")
+        other -> raise AssertionError, message: "No download detected. Could not find filename in #{inspect(other)}"
```
ftes, line 59: Should we support non-attachment downloads?
E.g. open a PDF in a new tab (without triggering the browser "save" dialog).

If so, how to differentiate attachment and non-attachment downloads?
- not at all
- optional argument: `assert_download("elixir.jpg", target: :attachment)`
- separate API function for non-attachment downloads: `assert_file("image/jpeg")`

lib/phoenix_test/assertions.ex
```diff
@@ -188,6 +188,23 @@ defmodule PhoenixTest.Assertions do
     session
   end
 
+  def assert_download(session, _) when not is_struct(session, PhoenixTest.Static) do
+    raise ArgumentError, "Only downloads via Phoenix.Controller are supported."
+  end
```
germsvel, line 193: Is there anything we can do for downloads that happen via LiveViews? Do you know what it might take to support that?

ftes: I think this is an unexpected case.

If you want to download files from within a liveview, I think you have to revert to JS (`URL.createObjectURL`).

Otherwise, the only option is to download via a normal link/redirect to a static controller based download.

In both cases, the `Live` driver is not involved.

germsvel:
Closing in favor of #305 . Thanks @ftes!

Issue #112, opened 2024-09-02
Handle input/select `phx-click` outside of forms


germsvel:
So far, our inputs and select dropdowns work _inside_ forms. 

But in LiveView, it's possible for them to send events to the server outside of forms with `phx-click` attributes. 

- [ ] checkboxes (PR #109)
- [x] select's options
- [x] radio buttons

Note: I originally thought we might be able to handle `phx-change` on text inputs, but it seems like [that's a form-only event handler](https://hexdocs.pm/phoenix_live_view/html-attrs.html#form-event-handlers)

alexslade:
@germsvel LiveView also supports phx-change on the select itself, without a form. 

Would you be open to a PR that supports this? I've had a poke around, looks like I might have to dig into ActiveForm tracking, to handle cases like this. 

If you've got any quick insights let me know, and I can have a run at it.

PR #116, opened 2024-09-05
Check/uncheck work with phx-click outside forms

germsvel:
Supersedes https://github.com/germsvel/phoenix_test/pull/109 
Resolves https://github.com/germsvel/phoenix_test/issues/112

What changed?
=============

We update `check` and `uncheck` to work with `phx-click` attributes
outside of forms.

The `check` case is straightforward. The `uncheck` is more complicated.

For the `uncheck` case, we first look for an checkbox that is visible.
That's contrary to what we were doing with form checkboxes -- where we
need a hidden input -- because a single checkbox can have a `phx-click`
for checking and unchecking. That removes the need for the hidden input
(like forms need).

So, if we find the checkbox and it has a `phx-click` attached, we
trigger that. If not, we check if the field belongs to a form. If so, we
find the hidden "uncheckbox" to get the value.

Then, we run into the problem that Phoenix's `phx-click` behavior on a
checkbox outside of a form makes sense but `render_click` behaves
differently:

- If a checkbox is getting checked, the `value` is included in the
  params
- If a checkbox is getting un-checked, the `value` is not included in
  the params (i.e. we get an empty map)

But `view |> element |> render_click` always includes the `value` of the
element in the payload. To bypass that, we have to avoid using the
`element` helper and directly send the event to `render_click/3` along
with an empty map for a payload.

### DIFF
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -126,17 +126,50 @@ defmodule PhoenixTest.Live do
   end
 
   def check(session, label) do
-    session
-    |> render_html()
-    |> Field.find_checkbox!(label)
-    |> then(&fill_in_field_data(session, &1))
+    field =
+      session
+      |> render_html()
+      |> Field.find_checkbox!(label)
+
+    cond do
+      Field.phx_click?(field) ->
+        session.view
+        |> element(field.selector)
+        |> render_click()
+        |> maybe_redirect(session)
+
+      Field.belongs_to_form?(field) ->
+        fill_in_field_data(session, field)
+
+      true ->
+        raise ArgumentError, """
+        Expected checkbox with selector #{inspect(field.selector)} to have a `phx-click` attribute or belong to a `form` element.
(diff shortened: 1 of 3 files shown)


lib/phoenix_test/live.ex
```diff
@@ -126,17 +126,50 @@ defmodule PhoenixTest.Live do
...
+    html = render_html(session)
+    field = Field.find_checkbox!(html, label)
+
+    cond do
+      Field.phx_click?(field) ->
+        event = Html.attribute(field.parsed, "phx-click")
+
+        session.view
+        |> render_click(event, %{})
+        |> maybe_redirect(session)
```
germsvel, line 161: This is intentional. Instead of using `element/2 |> render_click/1` we have to use `render_click/3`, passing the event and payload directly, so that we can pass an empty payload. Otherwise, the value is always included in the payload -- something that doesn't happen when unclicking a checkbox in Phoenix.

Issue #115, opened 2024-09-04
Select on nested form raises ArgumentError


rgraff:
Using `select(conn, "Users", from: "Metric")` will error on fields generated with `<inputs_for .../>`

 ```elixir
      ** (ArgumentError) value for hidden "automation[trigger][_touched]" must be one of ["_form_type,_persistent_id,_touched,name,type"], got: "name,type"
     code: |> select(metric.name, from: "Metric")
     stacktrace:
       (phoenix_live_view 0.20.17) lib/phoenix_live_view/test/live_view_test.ex:1102: Phoenix.LiveViewTest.call/2
       (phoenix_test 0.3.1) lib/phoenix_test/live.ex:219: PhoenixTest.Live.fill_form/3
```

It works with `<inputs_for skip_hidden={true} .../>`

germsvel:
Thanks for opening the issue @rgraff! 

Any chance you could include some code that makes this easy to reproduce?

rgraff:
Unfortunately, I can't reproduce the error in the test suite. I'm not exactly sure what happens with `<.inputs_for .../>` so I'm not confident my test setup is correct.

germsvel:
@rgraff can you use reproduce the error in a sample Phoenix app that you can share with us? Would that be easier than trying to add a test in `phoenix_test`?

rgraff:
@germsvel yes, I will do that in the next couple weeks. Have a deadline right now and have worked-around this issue for now.

germsvel:
> have worked-around this issue for now.

Excellent! 👍 Glad to hear it's not blocking you

germsvel:
@rgraff I'm going to go ahead and close this issue (trying to do some tidying up). If this is a problem in the future, just let me know. We can reopen this or open a new issue. Who knows, maybe newer versions of PhoenixTest already fixed this 😄

PR #117, opened 2024-09-05
Nested form with hidden inputs raises ArgumentError

rgraff:
Attempting to recreate the error I'm getting in my app for issue #115. I wrote a test that is a pretty good reproduction but does not fail.  😞


germsvel:
@rgraff if you don't mind, I'm going to close this PR since we cannot reproduce the issue. If you find a way to reproduce it, I'd be happy to give it a look. But I don't know how to address this otherwise.

PR #118, opened 2024-09-05
Find select option via 'exact=true' (breaking change)

ftes:
Find select option by exact match. Allows disambiguating options that share a common substring.

E.g.
```html
<select>
  <option value="orc">Orc</option>
  <option value="other_orc">Other Orc</option>
</select>
```

Currently, there is no way to distinguish `Orc` and `Other Orc` (`Found more than one element`).

### DIFF
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -101,7 +101,9 @@ defmodule PhoenixTest.Live do
     |> then(&fill_in_field_data(session, &1))
   end
 
-  def select(session, option, [{:from, label} | opts]) do
+  def select(session, option, opts) do
+    label = Keyword.fetch!(opts, :from)
+
     field =
       session
       |> render_html()
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -101,11 +101,11 @@ defmodule PhoenixTest.Live do
     |> then(&fill_in_field_data(session, &1))
   end
 
-  def select(session, option, from: label) do
+  def select(session, option, [{:from, label} | opts]) do
     field =
       session
       |> render_html()
-      |> Select.find_select_option!(label, option)
+      |> Select.find_select_option!(label, option, opts)
 
     cond do
       Select.belongs_to_form?(field) ->
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -101,11 +101,11 @@ defmodule PhoenixTest.Live do
     |> then(&fill_in_field_data(session, &1))
   end
 
-  def select(session, option, [{:from, label} | opts]) do
+  def select(session, option, from: label) do
     field =
       session
       |> render_html()
-      |> Select.find_select_option!(label, option, opts)
+      |> Select.find_select_option!(label, option)
 
     cond do
       Select.belongs_to_form?(field) ->
--- a/lib/phoenix_test/static.ex
+++ b/lib/phoenix_test/static.ex
@@ -117,10 +117,10 @@ defmodule PhoenixTest.Static do
     |> then(&fill_in_field_data(session, &1))
   end
 
-  def select(session, option, [{:from, label} | opts]) do
+  def select(session, option, from: label) do
     session
     |> render_html()
-    |> Select.find_select_option!(label, option, opts)
+    |> Select.find_select_option!(label, option)
     |> then(&fill_in_field_data(session, &1))
   end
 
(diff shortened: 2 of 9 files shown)


ftes:
Should `exact: true` be the default?
Since inputs are found by exact match on their label, that would be consistent.

germsvel:
> Should exact: true be the default?
> Since inputs are found by exact match on their label, that would be consistent.

Yeah, that's excellent question. And thanks for adding this. The need for `exact: true` makes total sense. 

My _guess_ is that we should make `exact: true` the default for `select`. That's my guess based on the "principle of least surprise". It's kind of gut feeling, but I always ask myself, "what would a user expect to be the default here?"

In this case, I think we typically write the full label name -- so we're _typically_ trying to do exact matches. That's different from assertions, where I think we're typically trying to do an inexact match (which is why the default there is `false`)

I think it's okay that it's a breaking change. I try to avoid them, but I think this is a good reason for having one.

ftes:
> My guess is that we should make exact: true the default for select. That's my guess based on the "principle of least surprise". It's kind of gut feeling, but I always ask myself, "what would a user expect to be the default here?"
> In this case, I think we typically write the full label name -- so we're typically trying to do exact matches. That's different from assertions, where I think we're typically trying to do an inexact match (which is why the default there is false)

After giving it some thought, I agree.
`exact=true` is a good default for finding select options.
I updated the PR and also removed the option again - I think it should be fine to keep things simple and non-configurable for now.

germsvel:
> I ... also removed the option again - I think it should be fine to keep things simple and non-configurable for now.

@ftes oh no! I hope my previous comment didn't make you remove the configurable `exact` flag. I thought that was a good portion (and I liked your other PR introducing that too). 

Since we're creating the breaking change here, it'd be nice to provide the `exact` option too (when we release this change), so that the upgrade path is fairly straightforward. If people want to keep the current behavior, then they can specify `exact: false`. I wouldn't want to release this change, break things for people, and not have a way for them to overcome those issues. 

We don't have to introduce that `exact` flag here, but we should release it as part of the same set of changes. So, I'm gonna hold off on merging this until we have those other changes lined up too. If you feel like reopening your other PR and reintroducing them here, that'd be great. Otherwise, I can add those when I'm done working on the form-helpers-that-accepts-selectors issue.

ftes:
Then I misunderstood you 😆  No problem, I'll re-add in this or another PR.

ftes:
`exact` option re-added.

lib/phoenix_test/live.ex
```diff
@@ -101,11 +101,11 @@ defmodule PhoenixTest.Live do
     |> then(&fill_in_field_data(session, &1))
   end
 
-  def select(session, option, from: label) do
+  def select(session, option, [{:from, label} | opts]) do
```
germsvel, line 104: When there's a single keyword arg, I sometimes pattern match like I did there with `from: label`. But if there's multiple, I think we should just call this `opts` and do a `Keyword.fetch!` inside the body of the function. 

I think that's a better approach because the keyword args are position dependent -- meaning if someone types: 

```elixir
session
|> select("Orc", exact: false, from: "Race")
```

this will error. 

What do you think? Make sense to move `Keyword.get` into the body of the function to pull out the label?

lib/phoenix_test/static.ex
```diff
@@ -117,10 +117,10 @@ defmodule PhoenixTest.Static do
     |> then(&fill_in_field_data(session, &1))
   end
 
-  def select(session, option, from: label) do
+  def select(session, option, [{:from, label} | opts]) do
```
germsvel, line 120: Same comment as in the `Live` implementation. 

This is probably my fault for using `from: label` in pattern matching, but since keyword args are order-dependent, what do you think about just calling this `opts` and pulling the `label` out in the body of the function with `Keyword.fetch!` (so it raises if we people don't pass it)?

germsvel (commented):
Looks good! Thanks for reintroducing the `exact` option. I had a couple of comments, but otherwise this is good to go.

lib/phoenix_test/live.ex
ftes: Could have used `Keyword.pop!` instead I guess, to prevent passing the `from` opt further along.
Not that it matters too much.

germsvel: Oh yeah, `pop!` would be nice. I suppose we could also go through and add a `Keyword.validate!` at some point to make sure we're only passing the correct options. I can do a follow up at some point to do that.

germsvel:
Opened #127 to keep track of this work. Thanks again for getting this in!

PR #119, opened 2024-09-10
Add playwright driver

ftes:
To get discussion kicked off.


lib/phoenix_test.ex
```diff
@@ -199,6 +199,11 @@ defmodule PhoenixTest do
   alias PhoenixTest.Driver
 
   @endpoint Application.compile_env(:phoenix_test, :endpoint)
+
+  def with_playwright(%Plug.Conn{} = conn, browser \\ :chromium) do
```
ftes, line 203: Which browser should we run with?
- customise: global config, per-test options
  - e.g. user configures default browser globally, and can override per test
- allow running single test with 1..3 browsers

lib/phoenix_test/playwright.ex
```diff
@@ -0,0 +1,150 @@
...
+  alias PhoenixTest.Html
+  alias PhoenixTest.OpenBrowser
+  alias PhoenixTest.Query
+
+  @endpoint Application.compile_env(:phoenix_test, :endpoint)
+
+  defstruct page: nil, within: :none, last_field_query: :none
+
+  def build(_conn, path, browser_name) do
+    {:ok, browser} = Playwright.launch(browser_name)
```
ftes, line 14: Not using `PlaywrightTest.Case`.
Which means we are missing teardown of the browser atm.

test/phoenix_test/static_test.exs
```diff
@@ -4,12 +4,14 @@ defmodule PhoenixTest.StaticTest do
...
-    %{conn: Phoenix.ConnTest.build_conn()}
+  setup context do
+    conn = Phoenix.ConnTest.build_conn()
+    conn = if context[:playwright], do: with_playwright(conn), else: conn
+    %{conn: conn}
   end
 
   describe "render_page_title/1" do
-    test "renders the page title", %{conn: conn} do
+    test_also_with_playwright "renders the page title", %{conn: conn} do
```
ftes, line 14: Working test 🥳

mix.exs
```diff
@@ -46,9 +46,11 @@ defmodule PhoenixTest.MixProject do
       {:jason, "~> 1.4"},
       {:phoenix, "~> 1.7.10"},
       {:phoenix_live_view, "~> 0.20.1"},
-      {:esbuild, "~> 0.8", only: :test, runtime: false},
+      {:esbuild, "~> 0.8", runtime: false},
```
ftes, line 49: `playwright` forces `esbuild` not only for test.

ftes:
Wdyt: Is the time ripe to begin work on this @germsvel?

germsvel:
> Wdyt: Is the time ripe to begin work on this @germsvel?

Thanks for asking @ftes. I have a branch that does some of the same things you do here, though some are different. I didn't open a draft PR because I don't want to give the impression that this is somehow almost ready. But the work is happening! Just gotta hang in there for me. Spending as much time as I can on it. 

That being said, happy to leave this draft open if it helps you. But I'm still exploring how best to introduce this separately.

ftes:
I'll just ahead and close this then 👍

PR #120, opened 2024-09-11
Check/uncheck with partial match via 'exact: false'

ftes:
Add `exact` option to `check/3` and `uncheck/3` to allow (un-)checking by substring match.

Related to #118.

To do
- [ ] consistent defaults for `exact` option - currently as a user of `phoenix_test` I don't know what to expect
- [ ] document `exact` option for `check/uncheck`


ftes:
>  I think it makes sense for the default to be exact: true in these cases.

Agreed, closing this

PR #123, opened 2024-09-16
Fix: Don't error on input without name attribute

ftes:
Given

```html
<input value="123" />
```

submit the form without error.
Just omit the input without a name attribute.

### DIFF
--- a/lib/phoenix_test/form.ex
+++ b/lib/phoenix_test/form.ex
@@ -91,14 +91,17 @@ defmodule PhoenixTest.Form do
   @pre_filled_default_text_inputs "input:not([disabled]):not([type])[value]"
 
   defp form_data(form) do
-    form_data(@hidden_inputs, form) ++
-      form_data(@checked_radio_buttons, form) ++
-      form_data(@checked_checkboxes, form) ++
-      form_data(@pre_filled_text_inputs, form) ++
-      form_data(@pre_filled_number_inputs, form) ++
-      form_data(@pre_filled_default_text_inputs, form) ++
-      form_data_textarea(form) ++
-      form_data_select(form)
+    Enum.reject(
+      form_data(@hidden_inputs, form) ++
+        form_data(@checked_radio_buttons, form) ++
+        form_data(@checked_checkboxes, form) ++
+        form_data(@pre_filled_text_inputs, form) ++
+        form_data(@pre_filled_number_inputs, form) ++
+        form_data(@pre_filled_default_text_inputs, form) ++
+        form_data_textarea(form) ++
+        form_data_select(form),
+      fn {key, _value} -> is_nil(key) end
+    )
   end
 
   defp form_data(selector, form) do
--- a/test/support/page_view.ex
+++ b/test/support/page_view.ex
@@ -213,6 +213,10 @@ defmodule PhoenixTest.PageView do
         <input type="checkbox" name="member_of_fellowship" />
       </div>
 
+      <label>
+        Ignored presentational input input <input value="123" />
+      </label>
+
       <button name="full_form_button" value="save">Save Full Form</button>
     </form>
 


germsvel:
@ftes thanks for opening this. I've been meaning to improve how we handle inputs without a `name` (although I wanted to do something different). 

I know when an `input` is missing a `name`, right now we're failing in a pretty ugly way. 

Rather than ignoring the input without a `name`, I wanted to make it error with a nice, helpful error telling the user that the input doesn't have a `name`.  That way, they can correct it. 

What do you think? Want to give that a try? Otherwise, we can close this, and I'll add that `name` validation later.

germsvel:
> Reading from the spec, I think a missing name should be supported ("if the attribute is specified")

Yeah, I think this may be one place where we want to deviate from the spec in order to be more helpful. I'm not 100% sold on it myself, but I _think_ that's the direction we'd want to go. 

I realize it's possible to have an input without a `name`. But is an input without a `name` useful? It's an essential part of the `name/value` combination that gets sent to the server when the input is submitted. 

Of course, with JS (including the `phx-hook` example), we can do all kinds of things and _not_ need a name. But the question I'm trying to ask myself is:

Which of these two is a worse _negative_ experience? 

- I create an `input` without a name because I don't need a name. The library gives me a nice error saying the `input` needs a `name`. My `input` doesn't need it b/c I'm using JS, but sure, I'll go ahead and add a name. OR
- I create an `input` without a name because I _forgot_ to put it there. But I need it for my form submission. The test keeps going without warning. Then, when I get to the assertion, something's not working. The record isn't being saved with the correct info. I go and debug, look at the params, and find that the data isn't being passed. Why? I look at the HTML, and I see the `input` there. It's fine. Why isn't it sending it? 

I think the second one is worse. The first one is an annoyance. The second one can lead to confusing results that require a lot of time debugging because it fails _silently_. 

So, that's my thought process right now.

ftes:
> The second one can lead to confusing results that require a lot of time debugging because it fails silently.

It's difficult to draw the line: What problems might users need assistance with, and which level of basic web dev literacy is required.

> My input doesn't need it b/c I'm using JS, but sure, I'll go ahead and add a name

Playing devil's advocate: Lets try out this new library and see how I can test my existing forms that are already battle proven. Oh, it's throwing a bunch of errors. But my forms are fine. Error: "Missing a name". Since when is that required? I wonder which other application code I will have to change for test purposes. Better skip the library.

I'm fine either way.

germsvel:
> It's difficult to draw the line: What problems might users need assistance with, and which level of basic web dev literacy is required.

It's not about dev literacy. Just about how easy it is to miss adding a `name` -- and if we silently ignore the input, it makes it hard to debug why things aren't working. 

I like the devil's advocate portion too. It's possible someone hates the library because of it, but on the other hand, it's nice if the library highlights potential errors in your code -- so the reaction might equally be, "wow, missed that input's `name`. Glad we added the library". 

So, naturally there's arguments on both sides. And it's a judgment call on my side.  I do hear your argument, so I'll think about it a bit more.

germsvel:
@ftes I've thought about this more, and I think I found a solution that is the best of both worlds -- but I'd love to hear your thoughts. 

- I think we _should not_ raise an error if an input is missing `name` when we're just loading the default form data (that's what I think you're doing in this PR, right?)
- But we _should_ raise a nice error (right now it raises a terrible error) if an input doesn't have a `name` when you're targeting that input with a `fill_in`, `select`, etc. 

So... we don't force you to do stuff on inputs that you're not targeting -- I agree that's annoying. Feels like PhoenixTest is trying to own too much. But we do try to help write a form input that works when you're targeting it with PhoenixTest. 

Thoughts?

ftes:
I agree fully.

> should not raise an error if an input is missing name when we're just loading the default form data (that's what I think you're doing in this PR, right?)

Yes, that's what this PR should achieve. I did not explain the intent well. I didn't have those two cases clearly separated in my head. It seems obvious now you stated it like that.

germsvel:
>  I didn't have those two cases clearly separated in my head.  

Neither did I! 😃 It took me a little bit of thinking (and looking at your PR) to realize that there's a difference.

PR #124, opened 2024-09-16
Add failing test: click_button without id attr

ftes:
Broken by 95e100b853ee33dda92b5279546e13d974ed11a6

### DIFF
--- a/test/phoenix_test/live_test.exs
+++ b/test/phoenix_test/live_test.exs
@@ -259,6 +259,13 @@ defmodule PhoenixTest.LiveTest do
         end)
       end
     end
+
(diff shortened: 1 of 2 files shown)


germsvel:
Thanks for opening this @ftes. I'll take a look at what's going on there.

germsvel:
@ftes I think this doesn't have anything to do with the button not having an `id` attribute. I think it's that the LiveView.JS is getting escaped weirdly, and we're using that as part of the selector to try to find the button. I still think we need to fix it. Just giving an update.

germsvel:
@ftes got this fixed on #125. Let me know if you think the problem was somewhere else.

PR #131, opened 2024-09-30
Copy headers across redirects

### LINKED ISSUE #126: Proposal: copy request headers from initial conn into redirect/navigate conn
We encounted a problem where the request headers set before the initial visit/2 are not present in subsequent redirects. Could there be a configuration that always copies request headers to the next request, or a way to inject a before redirect callback that could allow us to modify the conn ahead of time. Also there is an option for `Phoenix.ConnTest.recycle/2` that instructs which request headers to carry forward when recycling the conn, so it would be great to have a way to set that option. https://hexdocs.pm/phoenix/Phoenix.ConnTest.html#recycle/2

Linked is a gist using phoenix_playground that demonstrates a failure case.
https://gist.github.com/joeytrapp/35c984a67cf2eb6eb0f6f4b3ef9e060f

germsvel:
Resolves #126 

What changed?
============

In order to emulate browser behavior a little bit better, we want to copy headers across redirects.

Whenever we dispatch to the endpoint, the `Phoenix.ConnTest.recycle/2` automatically gets called. But by default it only recycles "accept", "accept-language", and "authorization".

This commit calls it manually copying _all_ request headers.

For more, see docs on
https://hexdocs.pm/phoenix/Phoenix.ConnTest.html#recycle/2

> Recycling receives a connection and returns a new connection,
containing cookies and relevant information from the given one.

> This emulates behaviour performed by browsers where cookies returned
in the response are available in following requests.

> By default, only the headers "accept", "accept-language", and
"authorization" are recycled. However, a custom set of headers can be specified by passing a list of strings representing its names as the second argument of the function.

> Note recycle/1 is automatically invoked when dispatching to the
endpoint, unless the connection has already been recycled.

## Need to check the following: 

- [x] Redirect on visit
- [x] Redirect when following links
- [x] Redirect when following form submissions

### DIFF
--- a/lib/phoenix_test.ex
+++ b/lib/phoenix_test.ex
@@ -214,17 +214,27 @@ defmodule PhoenixTest do
   you're visiting.
   """
   def visit(conn, path) do
-    case get(conn, path) do
+    conn
+    |> recycle(all_headers(conn))
+    |> get(path)
+    |> case do
       %{assigns: %{live_module: _}} = conn ->
         PhoenixTest.Live.build(conn)
 
       %{status: 302} = conn ->
         path = redirected_to(conn)
-        visit(conn, path)
+
+        conn
+        |> recycle(all_headers(conn))
+        |> visit(path)
 
       conn ->
         PhoenixTest.Static.build(conn)
     end
+  end
+
+  defp all_headers(conn) do
+    Enum.map(conn.req_headers, &elem(&1, 0))
   end
 
   @doc """
--- a/lib/phoenix_test/static.ex
+++ b/lib/phoenix_test/static.ex
@@ -239,9 +239,16 @@ defmodule PhoenixTest.Static do
   end
 
   defp perform_submit(session, form, form_data) do
-    session.conn
+    conn = session.conn
+
+    conn
+    |> recycle(all_headers(conn))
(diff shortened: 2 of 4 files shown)


germsvel:
I wanted to make sure it wasn't _unsafe_ to copy all the headers. 

For posterity's sake, I think we're okay following these rules (snatched them from a friend): 

- If POST 301 or POST 302, do not copy any headers
- If non-GET 303, do not copy any headers
- If redirecting to a different domain than the initial request, copy all headers EXCEPT Authorization
- Otherwise, copy all headers

Since we don't really handle 301s or 302s, I don't think we need to worry about that. I don't think people are redirecting to different domains (since we're _presumably_ testing internal apps), so I don't think the issue of `Authorization` is a big deal. 

So, I think we're okay copying all the headers for now. 

I also gave Capybara (a Ruby library that has rack-test which is similar to `Phoenix.ConnTest`) a look to see how they'd handled this in the past. I found this issue (https://github.com/teamcapybara/capybara/issues/464) which was resolved in this commit (https://github.com/teamcapybara/capybara/commit/33b1a27c3219ae2eeb1d9d3e325e7b104fe2a192) which I think basically agrees that it's fine to copy headers.

PR #135, opened 2024-10-05
Support phx-trigger-action

### LINKED ISSUE #47: Support `phx-trigger-action`

ftes:
Closes #47

### DIFF
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -339,7 +339,22 @@ defmodule PhoenixTest.Live do
   end
 
   defp maybe_redirect(html, session) when is_binary(html) do
-    maybe_put_patch_path(session)
+    case Form.find(html, "form[phx-trigger-action]") do
+      :not_found ->
+        maybe_put_patch_path(session)
+
+      {:found, form} ->
+        active_form = session.active_form
+        active_form? = form.selector == active_form.selector
+        form_data = form.form_data ++ if(active_form?, do: active_form.form_data, else: [])
+
+        session.conn
+        |> PhoenixTest.Static.build()
+        |> PhoenixTest.Static.submit_form(form.selector, form_data)
+
+      {:found_many, _} ->
+        raise raise ArgumentError, "Found multiple forms with phx-trigger-action."
+    end
   end
 
   defp maybe_put_patch_path(session) do
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -353,7 +353,7 @@ defmodule PhoenixTest.Live do
         |> PhoenixTest.Static.submit_form(form.selector, form_data)
 
       {:found_many, _} ->
-        raise raise ArgumentError, "Found multiple forms with phx-trigger-action."
+        raise ArgumentError, "Found multiple forms with phx-trigger-action."
     end
   end
 
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -50,10 +50,7 @@ defmodule PhoenixTest.Live do
     session.view
     |> element(selector, text)
     |> render_click()
-    |> begin_post_processing(session)
-    |> maybe_trigger_form_action()
-    |> maybe_redirect()
-    |> finish_post_processing()
+    |> maybe_redirect(session)
   end
 
   def click_button(session, selector, text) do
@@ -65,10 +62,7 @@ defmodule PhoenixTest.Live do
         session.view
         |> element(selector, text)
         |> render_click()
-        |> begin_post_processing(session)
-        |> maybe_trigger_form_action()
-        |> maybe_redirect()
-        |> finish_post_processing()
+        |> maybe_redirect(session)
 
       Button.belongs_to_form?(button) ->
         active_form = session.active_form
@@ -127,10 +121,7 @@ defmodule PhoenixTest.Live do
           session.view
           |> element(Select.select_option_selector(field, value))
           |> render_click()
-          |> begin_post_processing(session)
-          |> maybe_trigger_form_action()
-          |> maybe_redirect()
-          |> finish_post_processing()
+          |> maybe_redirect(session)
         end)
 
       true ->
@@ -151,10 +142,7 @@ defmodule PhoenixTest.Live do
         session.view
         |> element(field.selector)
         |> render_click()
-        |> begin_post_processing(session)
-        |> maybe_trigger_form_action()
-        |> maybe_redirect()
-        |> finish_post_processing()
+        |> maybe_redirect(session)
 
       Field.belongs_to_form?(field) ->
         fill_in_field_data(session, field)
@@ -176,10 +164,7 @@ defmodule PhoenixTest.Live do
 
         session.view
         |> render_click(event, %{})
-        |> begin_post_processing(session)
-        |> maybe_trigger_form_action()
-        |> maybe_redirect()
-        |> finish_post_processing()
+        |> maybe_redirect(session)
 
       Field.belongs_to_form?(field) ->
         html
@@ -204,10 +189,7 @@ defmodule PhoenixTest.Live do
         session.view
         |> element(field.selector)
         |> render_click()
-        |> begin_post_processing(session)
-        |> maybe_trigger_form_action()
-        |> maybe_redirect()
-        |> finish_post_processing()
+        |> maybe_redirect(session)
 
       Field.belongs_to_form?(field) ->
         fill_in_field_data(session, field)
@@ -242,10 +224,7 @@ defmodule PhoenixTest.Live do
     session.view
     |> file_input(form.selector, live_upload_name, [entry])
     |> render_upload(file_name)
-    |> begin_post_processing(session)
-    |> maybe_trigger_form_action()
-    |> maybe_redirect()
-    |> finish_post_processing()
+    |> maybe_redirect(session)
   end
 
   defp fill_in_field_data(session, field) do
@@ -271,10 +250,7 @@ defmodule PhoenixTest.Live do
       session.view
       |> form(form.selector, FormPayload.new(data_to_submit))
       |> render_change(additional_data)
-      |> begin_post_processing(session)
-      |> maybe_trigger_form_action()
-      |> maybe_redirect()
-      |> finish_post_processing()
+      |> maybe_redirect(session)
     else
       session
     end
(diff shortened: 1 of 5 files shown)


lib/phoenix_test/live.ex
```diff
@@ -330,7 +330,22 @@ defmodule PhoenixTest.Live do
   end
 
   defp maybe_redirect(html, session) when is_binary(html) do
-    maybe_put_patch_path(session)
+    case Form.find(html, "form[phx-trigger-action]") do
```
germsvel, line 333: I find it a little strange that we'd be looking for a form inside `maybe_redirect`. 

I imagine it'd be more duplication to do this in the places where we're submitting forms, but I wonder if we could find a better design for this?

I'd image there are places where we're dealing with a form (probably through a phx-change) where we already have the `form` data structure. Maybe we could check for a `Form.phx_trigger_action?` there? (similar to how we do it with other attributes like that)

germsvel (commented):
Thanks for taking a stab at this. I'm not sold on the implementation. I think it would work, but I'd like to see if we can find a better design so that it's not intertwined with the `maybe_redirect` logic. 

What do you think? Could you explore that?

lib/phoenix_test/live.ex
ftes: Actually, I think the `phx-trigger-action` is a bit like a client side redirect.

Two things to consider:
- The `phx-trigger-action` can be set on any re-render, not just when clicking a submit button for a form
- The `phx-trigger-action` could be set on any form, not necessarily the one the user is interacting with

I agree that the common use case is:
1. User submits form
2. `handle_event("save"` assigns `trigger_submit: true`
3. Browser performs HTTP POST

But I think we should limit the implementation to that common use case.

So instead I think the only alternative is:
Whenever something changed, check the HTML for `phx-trigger-action`.
If it's present on a form, then submit that form.

We could separate `maybe_redirect` and `maybe_trigger_ation`.
Do you think that would help?

ftes: Commit e2dd5b4b8aa40bfca4e332cdb45a33c9839b5d64

I separated them into two separate post processing steps.
I tried to keep both functions completely separate by introducing a token flow.

Feels over-engineered. Maybe just renaming `maybe_redirect` to `maybe_trigger_form_action_or_redirect` would be good enough?

germsvel: > Feels over-engineered. Maybe just renaming maybe_redirect to maybe_trigger_form_action_or_redirect would be good enough?

Yeah, I agree. I appreciate you taking the time to try a different approach to see if we landed at something better. But it does seem to introduce more confusion. I think your original `maybe_redirect` implementation was best -- especially since you explained how `trigger_action` works. I thought it was more limited. 

> I think its best to mimmick [LiveView JS](https://github.com/phoenixframework/phoenix_live_view/blob/main/assets/js/phoenix_live_view/dom_patch.js#L167) to support all kinds of uses cases:

I think that's a good north star for sure. As we evaluate what to include and not include, I think there are two levels that I like to keep in mind: 

1.  PhoenixTest _not_ lie to users -- so, tests shouldn't pass if production is going to break. 
2. It's less bad for tests to break when production would pass (e.g. maybe we don't have a complete implementation). Ideally, we can match production, but I think this should always be secondary to (1). 

All that being said, any chance to revert the last commit, and I think we might be good to go? Do you think it'd be helpful to write test cases for the scenarios you mentioned? (when a different event happens and the trigger-action still happens?)

ftes: Reverted and added tests.

Also stumbled over #143.

lib/phoenix_test/live.ex
```diff
@@ -339,7 +339,22 @@ defmodule PhoenixTest.Live do
...
+        active_form = session.active_form
+        active_form? = form.selector == active_form.selector
+        form_data = form.form_data ++ if(active_form?, do: active_form.form_data, else: [])
+
+        session.conn
+        |> PhoenixTest.Static.build()
+        |> PhoenixTest.Static.submit_form(form.selector, form_data)
+
+      {:found_many, _} ->
+        raise raise ArgumentError, "Found multiple forms with phx-trigger-action."
```
germsvel, line 356: There's a double `raise` here. Could we remove one of them? Are we testing this path? If not, can we add a test that covers this?

lib/phoenix_test/live.ex
```diff
@@ -339,7 +339,22 @@ defmodule PhoenixTest.Live do
...
+        maybe_put_patch_path(session)
+
+      {:found, form} ->
+        active_form = session.active_form
+        active_form? = form.selector == active_form.selector
+        form_data = form.form_data ++ if(active_form?, do: active_form.form_data, else: [])
+
+        session.conn
+        |> PhoenixTest.Static.build()
+        |> PhoenixTest.Static.submit_form(form.selector, form_data)
```
germsvel, line 353: I have a question here. Do you know what happens in a normal LiveView if a form has something like a `phx-change` that redirects you somewhere _and_ it has a `phx-trigger-action`? Which redirect "wins out"? 

In this case, we're giving "priority" to the `phx-trigger-action`, which I imagine is the right call since most times I'd expect `phx-change` _not_ to be doing the redirecting. But I'm curious if we know what happens in those scenarios (and if we should even concern ourselves with them, or if it's just an edge case that we shouldn't cover).

germsvel, line 353: I would _imagine_ our implementation here is good. But I haven't done stuff with `phx-trigger-action`, so want to ask. Do you know if we should be using [LiveViewTest's `follow_trigger_action` function](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveViewTest.html#follow_trigger_action/2) to get the method and data to send?

test/phoenix_test/live_test.exs
```diff
@@ -1048,6 +1048,22 @@ defmodule PhoenixTest.LiveTest do
       |> assert_has("#form-data", text: "hidden_race: hobbit")
     end
 
+    test "phx-trigger-action causes POST to static view", %{conn: conn} do
+      conn
+      |> visit("/live/index")
+      |> fill_in("Trigger action", with: "engage")
+      |> submit()
+      |> assert_has("#form-data", text: "hidden: included")
+      |> assert_has("#form-data", text: "trigger_action: engage")
```
germsvel, line 1057: 🎉

test/support/index_live.ex
```diff
@@ -441,6 +441,19 @@ defmodule PhoenixTest.IndexLive do
...
     <div id="hook-with-redirect" phx-hook="SomeOtherHook"></div>
+
+    <form
+      id="trigger-form"
+      phx-submit="trigger-form"
+      phx-trigger-action={@trigger_submit}
+      action="/page/create_record"
+      method="post"
+    >
+      <input type="hidden" name="hidden" value="included" />
```
germsvel, line 452: Could we name this hidden input something more descriptive so it's easier to see in the tests that the trigger-action is being performed? 

Maybe something like `name="trigger-action-input-name"` and the value `value="trigger-action-value"` would make it more obvious?

test/support/index_live.ex
```diff
@@ -441,6 +441,19 @@ defmodule PhoenixTest.IndexLive do
...
+      <label>Trigger action <input type="text" name="trigger_action" /></label>
+    </form>
+
+    <button phx-click="trigger-form">Trigger from elsewhere</button>
```
germsvel, line 456: 👍

germsvel (commented):
Thanks again for doing this! 

I have a few comments and questions that I hope are helpful. Let me know if anything is unclear. I can also try to help out with implementation more, but it seems like you're more up to speed with how `phx-trigger-action` works in LiveView.

lib/phoenix_test/live.ex
ftes: Ouch, good catch. Adding test.

lib/phoenix_test/live.ex
ftes: Looking at the `follow_trigger_action`, we should be fine.
It does the same things we're doing (ensure `phx-trigger-action` attr present, then use `method` and `action` to HTTP submit the form).

## Should we use `follow_trigger_action` anyway?

Maybe. 

In future the implementations might drift apart (theoretically, not saying it's likely).
Using `follow_trigger_action` under the hood would prevent that. 

## No, rather be consistent with how we submit static forms from LiveViews

We have a precedent in PhoenixTest: static form submits from a LiveView (without phx-trigger-action).

https://github.com/germsvel/phoenix_test/blob/d62c264cd2fc1903f19c8edf902824e83589b834/lib/phoenix_test/live.ex#L299-L302

For that we _could_ use [`LiveViewTest.submit_form`](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveViewTest.html#submit_form/2). Which uses the same [helper function](https://github.com/phoenixframework/phoenix_live_view/blob/v0.20.17/lib/phoenix_live_view/test/live_view_test.ex#L1823C9-L1823C56) as `LiveViewTest.follow_trigger_action`.

For the sake of simplicity (my interpretation) we don't do that and rather just call the Static driver directly.

ftes: I think you maybe have misread the code. Or I'm wrong. Either way, I'm adding a test to clarify.

My understanding:
This implementation already prioritizes redirects or patches.
`phx-trigger-action` is only followed if no patch or redirect happened.
I'm pretty sure this is also what LiveView does.

These two clauses match first:

```ex
defp maybe_redirect({:error, {:redirect, %{to: path}}}, session) do
defp maybe_redirect({:error, {:live_redirect, %{to: path}}} = result, session) do
```

So, given
```heex
<form phx-trigger-action={@trigger_submit} phx-change="redirect">
```

```ex
handle_event("redirect", _, socket) do
  socket
  |> push_navigate(to: "/elsewhere")
  |> assign(:trigger_submit: true)
end
```

a change to the form will cause a redirect to `/elsewhere`. And ignore the `phx-trigger-action`.

ftes: Double checked the actual behaviour:
- `redirect`: wins over phx-trigger-action (✅  this PR is ok)
- `navigate`: wins over phx-trigger-action (✅  this PR is ok)
- `patch`: is executed before phx-trigger-action, both are applied (✅  this PR is ok)

So I think we're fine.

You're right in that `phx-trigger-action` is given priority over `maybe_put_patch_path`.
But if there's a `phx-trigger-action`, there's no point in updating the `session.current_path`, because we're about to POST to another path anyway.

ftes: Once again, running the same test against an actual browser as well as the `Live` driver would give peace of mind here :).

germsvel (approved):
Just now getting back to this. Thanks for all the work! Looks good 👍

PR #136, opened 2024-10-06
Add Playwright driver

ftes:
@germsvel I know you're working on this anyway. I'm just leaving this here in case you want to have a look.

I wanted to play around with playwright a bit to get more input for my upcoming CodeBEAM talk.
Fairly smooth sailing so far. I see you've been active in `playwright-elixir` getting things ready 🙂 .


test/support/page_view.ex
```diff
@@ -66,13 +66,13 @@ defmodule PhoenixTest.PageView do
       data-to="/page/delete_record"
       data-csrf="sometoken"
     >
-      Data-method Delete
+      Data_method Delete
```
ftes, line 69: If `exact: false` playwright is also case insensitive. While PhoenixTest is case sensitive.

lib/phoenix_test/playwright.ex
```diff
@@ -0,0 +1,264 @@
...
+
+      session
+    end
+
+    defp found?(session, selector, opts, query_attrs \\ []) do
+      # TODO support: count, at
+      css_selector =
+        case selector do
+          css when is_binary(css) -> css
+          other -> PhoenixTest.Locators.compile(other, render_html(session))
```
ftes, line 41: Instead of using the `Locator.compile`, we should compile directly to a Playwright locator.
Making use of the `internal:label` engine.

Then we also don't have to render the HTML. Which does not include automatic retry.

config/runtime.exs
```diff
@@ -0,0 +1,3 @@
+import Config
+
+config :playwright, LaunchOptions, headless: String.to_atom(System.get_env("PLAYWRIGHT_HEADLESS", "true")) != false
```
ftes, line 3: I think think this can be removed.

It's easier to do this in the test via

```ex
use PhoenixTest.Case,
  playwright: :chromium,
  headless: false,
  slow_mo: 500
```

lib/phoenix_test/phoenix_test_case.ex
```diff
@@ -0,0 +1,66 @@
...
+      browser = setup_browser(client, runner_options, launch_options)
+
+      [browser: browser]
+    end
+
+    def setup_playwright(%{playwright: _} = context) do
+      page = Playwright.Browser.new_page(context.browser)
+      ExUnit.Callbacks.on_exit(fn -> Playwright.Page.close(page) end)
+
+      [conn: page]
```
ftes, line 41: This just overwrites the `conn` context field with a playwright `page`.

Probably not the best way of doing this.
Just chose this as it was minimally invasive.

lib/phoenix_test/phoenix_test_case.ex
```diff
@@ -0,0 +1,66 @@
+defmodule PhoenixTest.Case do
```
ftes, line 1: Mostly copied from [PlaywrightTest.Case](https://github.com/mechanical-orchard/playwright-elixir/blob/main/lib/playwright_test/case.ex)

test/phoenix_test/live_test.exs
```diff
@@ -1001,21 +1047,27 @@ defmodule PhoenixTest.LiveTest do
...
 
   describe "shared form helpers behavior" do
-    test "triggers phx-change validations", %{conn: conn} do
+    test_also_with_playwright "triggers phx-change validations", %{conn: conn} do
       conn
       |> visit("/live/index")
-      |> fill_in("Email", with: nil)
+      |> within("#email-form", fn session ->
+        session
+        |> fill_in("Email", with: "email")
```
ftes, line 1055: Add text first, them remove.
Otherwise this is a no-op when using playwright.

test/phoenix_test/live_test.exs
```diff
@@ -196,15 +201,19 @@ defmodule PhoenixTest.LiveTest do
       |> refute_has("#form-data", text: "disabled_textarea:")
     end
 
-    test "can click button that does not submit form after filling form", %{conn: conn} do
+    test_also_with_playwright "can click button that does not submit form after filling form", %{conn: conn} do
       conn
       |> visit("/live/index")
-      |> fill_in("Email", with: "some@example.com")
+      |> within("#email-form", fn session ->
```
ftes, line 207: Workaround because these are not being treated equally by `Live` and `Static` drivers due to bug in `Query` (second input is ignored):

```html
<label for="email">Email</label>
<input id="email" name="email" />

<label>Email<input type="email" name="email" /></label>
```

config/runtime.exs
ftes: Removed

.github/workflows/ci.yml
```diff
@@ -69,5 +69,8 @@ jobs:
     - name: Check Formatting
       run: mix format --check-formatted
 
+    - name: Install playwright browsers
+      run: mix playwright.install
```
ftes, line 73: Cache?

lib/phoenix_test/phoenix_test_case.ex
```diff
@@ -0,0 +1,66 @@
...
+  @moduledoc false
```
ftes, line 2: Add docs (API tbd)


```ex
use PhoenixTest.Case,
  playwright: chromium,
  headless: false,
  slow_mo: 500
```

ftes:
`mix playwright.install` made set up in CI easy :) No caching yet tho.

ftes:
I don't think we need to use much of what `playwright-elixir` offers (or will offer in the future).

Idea: Talk directly to the playwright server. Implement only the relevant parts of the Websocket protocol.

Maybe I'm being naive and `playwright-elixir` is already doing more heavy lifting relevant for this use case than I think.
That would leave us somewhere inbetween the two options

germsvel:
> Talk directly to the playwright server. Implement only the relevant parts of the Websocket protocol.
Maybe I'm being naive and playwright-elixir is already doing more heavy lifting relevant for this use case than I think.

> Somewhat comparable to cuprite. Buth with playwright (server) as an additional abstraction (component) inbetween, instead of raw CDP.

That's exactly where I landed too. I don't know if I'm misrepresenting what `playwright-elixir` is doing for us, but I like the idea of `cuprite`. 

I started exploring that (even started creating a project to handle Elixir <-> CDP communication through websockets), but had to halt due to time constraints. 

So... now, I don't know if we need Wallaby or Playwright. It would be _cool_ (but not sure if a ton more work) to create an Elixir CDP wrapper that allows us to communicate with chrome (kind of like `ferrum` which `cuprite` uses). And then PhoenixTest just uses that.

ftes:
Elixir CDP wrapper: [bitcrowd/chromic_pdf](https://github.com/bitcrowd/chromic_pdf) might be a good starting point.

But I still think "vanilla" playwright (not via `playwright-elixir`) might also be a good option.
Higher-level abstraction (less for us to implement?), multi browser support out of the box, tooling (e.g. installing browsers).

germsvel:
> Elixir CDP wrapper: [bitcrowd/chromic_pdf](https://github.com/bitcrowd/chromic_pdf) might be a good starting point.

Oh cool. I didn't know about that package. I'll take a look. 

> But I still think "vanilla" playwright (not via playwright-elixir) might also be a good option.
> Higher-level abstraction (less for us to implement?), multi browser support out of the box, tooling (e.g. installing browsers).

Yeah, that's an interesting idea. I didn't realize that's what you had suggested at first (though, now I see it 😄 ). I like that idea too. I can see how we need maybe 1/10th of what playwright-elixir is trying to do. We could do it directly with playwright, and leverage what we already have in `phoenix_test` the way you have been thinking about implementing Wallaby and Playwright-Elixir (meaning, we rely on `render_html` a lot and then do assertions and stuff like that on our side)

ftes:
Closing in favour of #145

ftes:
Closing in favour of #145

PR #138, opened 2024-10-09
Fix: Latest HTML position trumps latest entry

ftes:
Looking at #137 I realised we probably have a bug (already prior to #137).

## Todo
- [x] Add failing test
- [ ] Fix

## Given
```html
<label>Email duplicate 1<input id="email-duplicate-1" name="email" /></label>
<label>Email duplicate 2<input id="email-duplicate-2" name="email" /></label>
```

```ex
|> fill_in("Email duplicate 2", with: "Latest HTML position")
|> fill_in("Email duplicate 1", with: "Latest fill_in")
```

## Expected
```ex
# POST email=Latest fill_in&email=Latest HTML position
params = %{"email" => "Latest HTML position"}
```

## Actual
```ex
params = %{"email" => "Latest fill_in"}
```

https://www.w3.org/TR/2011/WD-html5-20110525/association-of-controls-and-forms.html#constructing-form-data-set

### DIFF
--- a/test/phoenix_test/static_test.exs
+++ b/test/phoenix_test/static_test.exs
@@ -828,6 +828,15 @@ defmodule PhoenixTest.StaticTest do
       |> assert_has("h1", text: "Record deleted")
     end
 
+    test "submits input value with latest HTML position, not latest fill_in", %{conn: conn} do
+      conn
+      |> visit("/page/index")
+      |> fill_in("Email duplicate 2", with: "Latest HTML position")
+      |> fill_in("Email duplicate 1", with: "Latest fill_in")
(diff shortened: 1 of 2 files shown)


ftes:
Unrelated, but a recurring thought:

Adding a JS driver will help prevent some of these HTML-semantics-bugs.
Because running the existing tests with an actual browser shows that at least for those cases we are HTML-spec-conformant.

Along the same lines I've found a few edge case bugs while working on #136  and #74.

ftes:
I'll close this for now. Not too common a problem, no easy fix.
But maybe something for back of mind.

PR #141, opened 2024-10-24
Fix: type=check should be type=checkbox

ftes:
Found via #136 

- fails when running with an actual browser (playwright)


germsvel (approved):
Nice catch. Thanks for doing this!

PR #142, opened 2024-10-24
Fix: Raise error if implicit + explicit label found

ftes:
Found via #136 

## Given
```html
<label for="greeting">Hello</label>
<input id="greeting" />

<label>Hello <input id="second_greeting" /></label>
```

```ex
session
|> fill_in("Hello", with: "world")
```

## Expected
```ex
ArgumentError: Found many elements with label "Hello" and matching the provided selectors
```

## Actual
No error. Label with explicit `for` wins, other one ignored.

### DIFF
--- a/test/phoenix_test/live_test.exs
+++ b/test/phoenix_test/live_test.exs
@@ -1040,8 +1040,8 @@ defmodule PhoenixTest.LiveTest do
       |> visit("/live/index")
       |> within("#changes-hidden-input-form", fn session ->
         session
-        |> fill_in("Name", with: "Frodo")
-        |> fill_in("Email", with: "frodo@example.com")
+        |> fill_in("Name for hidden", with: "Frodo")
+        |> fill_in("Email for hidden", with: "frodo@example.com")
       end)
       |> assert_has("#form-data", text: "name: Frodo")
       |> assert_has("#form-data", text: "email: frodo@example.com")
--- a/test/phoenix_test/query_test.exs
+++ b/test/phoenix_test/query_test.exs
@@ -453,6 +453,19 @@ defmodule PhoenixTest.QueryTest do
       end
     end
 
+    test "raises error if label has for attribute and nested input" do
+      html = """
+      <label for="other_greeting">Hello <input /></label>
+      <input id="other_greeting" />
+      """
+
+      msg = ~r/Found a label which references two different inputs/
+
+      assert_raise ArgumentError, msg, fn ->
+        Query.find_by_label!(html, "input", "Hello")
+      end
+    end
+
     test "returns found element" do
       html = """
       <label for="greeting">Hello</label>
--- a/test/phoenix_test/query_test.exs
+++ b/test/phoenix_test/query_test.exs
@@ -438,6 +438,21 @@ defmodule PhoenixTest.QueryTest do
       end
     end
 
+    test "raises error if multiple labels and inputs match (one explicit, one implicit)" do
+      html = """
+      <label for="greeting">Hello</label>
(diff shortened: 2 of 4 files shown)


germsvel:
I like this idea! Do you know what HTML spec would do for this? Is this how it behaves, or does it target the implicit one (or neither one or both)?

ftes:
> Do you know what HTML spec would do for this? Is this how it behaves, or does it target the implicit one (or neither one or both)?

I think the HTML spec has no opinion on this.

Using labels to find input fields is a semantic layer that phoenix_test adds to reflect user behaviour.
The HTML spec (and browser) doesn't really care about the labels in the context of form submission.

Did I understand your question correctly?

germsvel:
> Did I understand your question correctly?

I'm not sure. 

My question was aimed at figuring out how the HTML spec would resolve the issue you're presenting in this PR: that of a label having a `for` that points to an element's `id` and _also_ having an implicit input. 

I wasn't sure, but I was curious if HTML spec might have some info as to which element it would label. 

Reading [the spec](https://html.spec.whatwg.org/multipage/forms.html#the-label-element): 

> If the [for](https://html.spec.whatwg.org/multipage/forms.html#attr-label-for) attribute is not specified, but the [label](https://html.spec.whatwg.org/multipage/forms.html#the-label-element) element has a [labelable element](https://html.spec.whatwg.org/multipage/forms.html#category-label) descendant, then the first such descendant in [tree order](https://dom.spec.whatwg.org/#concept-tree-order) is the [label](https://html.spec.whatwg.org/multipage/forms.html#the-label-element) element's [labeled control](https://html.spec.whatwg.org/multipage/forms.html#labeled-control).

It seems like the element with the `id` that matches the `for` would be the "labelable" element in that case. So, I think the solution you're presenting would be the correct behavior. 👍

test/phoenix_test/query_test.exs
```diff
@@ -438,6 +438,21 @@ defmodule PhoenixTest.QueryTest do
       end
     end
 
+    test "raises error if multiple labels and inputs match (one explicit, one implicit)" do
```
germsvel, line 441: 👍 thanks for adding a `query_test` for this!

test/phoenix_test/live_test.exs
```diff
@@ -1040,8 +1040,8 @@ defmodule PhoenixTest.LiveTest do
       |> visit("/live/index")
       |> within("#changes-hidden-input-form", fn session ->
         session
-        |> fill_in("Name", with: "Frodo")
-        |> fill_in("Email", with: "frodo@example.com")
+        |> fill_in("Name for hidden", with: "Frodo")
+        |> fill_in("Email for hidden", with: "frodo@example.com")
```
germsvel, line 1044: I'm nor sure I'm following -- why did we need this change?

germsvel (commented):
A small question, but otherwise looks good. 👍 

Thanks for handling this!

test/phoenix_test/live_test.exs
ftes: These labels were duplicated in the page.
That's what triggered this investigation :)

The duplication just previously went unnoticed due to the bug.
But now that the bug is fixed, this test would otherwise fail (with an error along the lines of "found multiple inputs with same label").

ftes:
> I still think it makes to raise an error here. We don't want people to have an explicit and an implicit label and us just choose the explicit one. So, I think it makes sense do include this PR.

I added a check and test for this. As you said, we're now stricter than the HTML spec.
But I agree that's a good thing to assist users in finding errors.

PR #144, opened 2024-10-26
Fix: Update current path on static submit

### LINKED ISSUE #143: Submitting static form does not update current path
When submitting a static form from a live view, the current path remains unchanged (stuck at the live view path).

Possibly also broken when submitting a form from a static view.

Found while working on #135.

ftes:
Closes #143


germsvel (approved):
Thanks!

PR #145, opened 2024-10-27
Add Playwright driver (no deps)

ftes:
Add `Playwright` test driver.
It communicates with the playwright `node.js` server via a `Port`.
No additional dependencies.

## Details
- bump to elixir `1.18` to use parameterized ExUnit tests (run live/static tests also with playwright)
- add minimal playwright protocol (communicate with playwright node.js server via `Port`)
- add `PhoenixTest.Playwright driver (very close to feature parity, some known inconsistencies that are documented via tests)
- add `PhoenixTest.Case` to allow easily switching drivers via `@tag :playwright`
- add optional deps `ecto` + `postgres` to test sandboxing of concurrent browser tests

## Out of scope (just document how to)
- installing: nodejs, npm package playwright, playwright browser(s)


lib/phoenix_test/playwright/connection.ex
```diff
@@ -0,0 +1,144 @@
...
+  ])
+
+  def start_link(recipient \\ self(), config) do
+    GenServer.start_link(__MODULE__, {recipient, config}, name: __MODULE__, timeout: 1000)
+  end
+
+  def launch_browser(type, opts \\ []) do
+    opts = Enum.into(opts, %{headless: true, slowMo: nil})
+
+    mapping =
```
ftes, line 23: Preserve mapping in Connection process.

lib/phoenix_test/playwright/connection.ex
```diff
@@ -0,0 +1,148 @@
...
+    :buffer,
+    msg_history: %{},
+    pending_responses: %{}
+  ])
+
+  def start_link(recipient \\ self(), config) do
+    GenServer.start_link(__MODULE__, {recipient, config}, name: __MODULE__, timeout: 1000)
+  end
+
+  def launch_browsers(types, opts \\ []) do
```
ftes, line 20: Todo: Launch on demand

test/test_helper.exs
```diff
@@ -2,3 +2,9 @@ ExUnit.start()
 
 {:ok, _} = Supervisor.start_link([{Phoenix.PubSub, name: PhoenixTest.PubSub}], strategy: :one_for_one)
 {:ok, _} = PhoenixTest.Endpoint.start_link()
+
+{:ok, _} = PhoenixTest.Playwright.Connection.start_link(%{})
```
ftes, line 6: Todo: Connect on demand (no overhead if not running any feature tests)

ftes:
In my eyes this is the most promising approach of the three so far (#136 , #74 #145).

The amount of code required to interface with playwright via a `Port` is fairly small.
And this gives us the flexibility to tailor the exact behaviour to `PhoenixTest`.

@germsvel If you have time I'd appreciate hearing your thoughts.

lib/phoenix_test/playwright.ex
```diff
@@ -0,0 +1,261 @@
...
+  defdelegate open_browser(session), to: Playwright
+  defdelegate open_browser(session, open_fun), to: Playwright
+  defdelegate unwrap(session, fun), to: Playwright
+  defdelegate current_path(session), to: Playwright
+
+  defdelegate assert_has(session, selector), to: Playwright
+  defdelegate assert_has(session, selector, opts), to: Playwright
+  defdelegate refute_has(session, selector), to: Playwright
+  defdelegate refute_has(session, selector, opts), to: Playwright
+  defdelegate assert_path(session, path), to: Assertions
```
ftes, line 257: Wrap in retry

test/phoenix_test/playwright_test.exs
```diff
@@ -0,0 +1,26 @@
+defmodule PhoenixTest.PlaywrightTest do
+  use PhoenixTest.Case, async: true
+
+  import PhoenixTest
+
+  @moduletag :playwright
```
ftes, line 6: Using a `moduletag` to enable playwright (as opposed to a `use` arg) in preparation for ExUnit 1.18.

This will allow us to run tests concurrently in multiple browsers using [parameterized tests](https://hexdocs.pm/ex_unit/main/ExUnit.Case.html#module-parameterized-tests):

```ex
use PhoenixTest.Case,
  async: true,
  parameterize: Enum.map([:chromium, :firefox], &%{playwright: &1})
```

test/phoenix_test/playwright_test.exs
```diff
@@ -0,0 +1,26 @@
...
+  describe "render_page_title/1" do
+    test "uses playwright driver by default", %{conn: conn} do
+      session = visit(conn, "/live/index")
+      assert %PhoenixTest.Playwright{} = session
+
+      title = PhoenixTest.Driver.render_page_title(session)
+      assert title == "PhoenixTest is the best!"
+    end
+
+    @tag playwright: false
```
ftes, line 17: Opt-out or opt-in?

I figure once you enable `playwright` for a test module, you probably want to run most tests within with playwright.

If not, you can always use `@describetag playwright: false` to disable for a group of tests.

test/phoenix_test/assertions_test.exs
```diff
@@ -8,17 +8,17 @@ defmodule PhoenixTest.AssertionsTest do
...
   describe "assert_has/2" do
-    test "succeeds if single element is found with CSS selector", %{conn: conn} do
+    also_test_js "succeeds if single element is found with CSS selector", %{conn: conn} do
       conn
       |> visit("/page/index")
       |> assert_has("[data-role='title']")
     end
 
+    # TODO Playwright: Fix error message
+    @tag playwright: false
```
ftes, line 21: Playwright is now opt-out.

Pro: Ensures we prefer running any new tests also against playwright.
Conn: Somewhat cryptic.

test/phoenix_test/assertions_test.exs
```diff
@@ -8,17 +8,17 @@ defmodule PhoenixTest.AssertionsTest do
...
   alias PhoenixTest.Live
 
-  setup do
-    %{conn: Phoenix.ConnTest.build_conn()}
-  end
+  @moduletag :playwright
 
   describe "assert_has/2" do
-    test "succeeds if single element is found with CSS selector", %{conn: conn} do
+    also_test_js "succeeds if single element is found with CSS selector", %{conn: conn} do
```
ftes, line 14: Replace macro with [parameterized tests](https://hexdocs.pm/ex_unit/main/ExUnit.Case.html#module-parameterized-tests) once Elixir 1.18 is out?

test/phoenix_test/live_test.exs
```diff
@@ -196,15 +206,19 @@ defmodule PhoenixTest.LiveTest do
       |> refute_has("#form-data", text: "disabled_textarea:")
     end
 
-    test "can click button that does not submit form after filling form", %{conn: conn} do
+    also_test_js "can click button that does not submit form after filling form", %{conn: conn} do
       conn
       |> visit("/live/index")
-      |> fill_in("Email", with: "some@example.com")
+      |> within("#email-form", fn session ->
```
ftes, line 212: Workaround for https://github.com/germsvel/phoenix_test/pull/142

test/phoenix_test/live_test.exs
```diff
@@ -314,14 +330,16 @@ defmodule PhoenixTest.LiveTest do
   end
 
   describe "fill_in/4" do
-    test "fills in a single text field based on the label", %{conn: conn} do
+    also_test_js "fills in a single text field based on the label", %{conn: conn} do
       conn
       |> visit("/live/index")
-      |> fill_in("Email", with: "someone@example.com")
-      |> assert_has(input(label: "Email", value: "someone@example.com"))
+      |> within("#email-form", fn session ->
```
ftes, line 336: Workaround for https://github.com/germsvel/phoenix_test/pull/142

test/phoenix_test/live_test.exs
```diff
@@ -370,24 +388,28 @@ defmodule PhoenixTest.LiveTest do
       |> assert_has("#form-data", text: "user:role: El Jefe")
     end
 
-    test "can be used to submit form", %{conn: conn} do
+    also_test_js "can be used to submit form", %{conn: conn} do
       conn
       |> visit("/live/index")
-      |> fill_in("Email", with: "someone@example.com")
+      |> within("#email-form", fn session ->
```
ftes, line 394: Workaround for https://github.com/germsvel/phoenix_test/pull/142

test/phoenix_test/live_test.exs
```diff
@@ -370,24 +388,28 @@ defmodule PhoenixTest.LiveTest do
...
       |> click_button("Save Email")
       |> assert_has("#form-data", text: "email: someone@example.com")
     end
 
-    test "can be combined with other forms' fill_ins (without pollution)", %{conn: conn} do
+    also_test_js "can be combined with other forms' fill_ins (without pollution)", %{conn: conn} do
       conn
       |> visit("/live/index")
-      |> fill_in("Email", with: "frodo@example.com")
+      |> within("#email-form", fn session ->
```
ftes, line 404: Workaround for https://github.com/germsvel/phoenix_test/pull/142

test/phoenix_test/live_test.exs
```diff
@@ -466,44 +491,27 @@ defmodule PhoenixTest.LiveTest do
       |> assert_has("#form-data", text: "[elf, dwarf]")
     end
 
-    test "works with phx-click outside of forms", %{conn: conn} do
```
ftes, line 469: `phx-click` on `option` is not actually supported

test/phoenix_test/live_test.exs
```diff
@@ -466,44 +491,27 @@ defmodule PhoenixTest.LiveTest do
...
-      |> assert_has("#form-data", text: "selected: [dog, cat]")
-    end
-
-    test "can target a label with exact: false", %{conn: conn} do
+    also_test_js "can target a label with exact: false", %{conn: conn} do
       conn
       |> visit("/live/index")
       |> within("#complex-labels", fn session ->
-        select(session, "Dog", from: "Choose a pet:", exact: false)
+        select(session, "Cat", from: "Choose a pet:", exact: false)
```
ftes, line 498: `Dog` is already initially selected. So "select"-ing it in a browser won't trigger `phx-change`.
Need to select a different option, i.e. `Cat`.

test/phoenix_test/live_test.exs
```diff
@@ -662,31 +673,31 @@ defmodule PhoenixTest.LiveTest do
...
       |> assert_has("#form-data", text: "contact: mail")
     end
 
-    test "works with a phx-click outside of a form", %{conn: conn} do
+    also_test_js "works with a phx-click outside of a form", %{conn: conn} do
       conn
       |> visit("/live/index")
       |> within("#not-a-form", fn session ->
-        choose(session, "Huey")
+        choose(session, "Dewey")
```
ftes, line 695: Choose something other than the initial selection.

test/phoenix_test/live_test.exs
```diff
@@ -1002,20 +1029,28 @@ defmodule PhoenixTest.LiveTest do
   end
 
   describe "shared form helpers behavior" do
-    test "triggers phx-change validations", %{conn: conn} do
+    also_test_js "triggers phx-change validations", %{conn: conn} do
       conn
       |> visit("/live/index")
-      |> fill_in("Email", with: nil)
+      |> within("#email-form", fn session ->
```
ftes, line 1035: Workaround for #142

test/phoenix_test/live_test.exs
```diff
@@ -1002,20 +1029,28 @@ defmodule PhoenixTest.LiveTest do
...
+        session
+        |> fill_in("Email", with: "email")
```
ftes, line 1037: To trigger a `phx-change`, the value needs to actually change.

test/phoenix_test/live_test.exs
```diff
@@ -1002,20 +1029,28 @@ defmodule PhoenixTest.LiveTest do
...
+      end)
       |> assert_has("#form-errors", text: "Errors present")
     end
 
-    test "sends _target with phx-change events", %{conn: conn} do
+    also_test_js "sends _target with phx-change events", %{conn: conn} do
       conn
       |> visit("/live/index")
-      |> fill_in("Email", with: "frodo@example.com")
+      |> within("#email-form", fn session ->
```
ftes, line 1046: Workaround for #142

test/support/index_live.ex
```diff
@@ -127,26 +127,26 @@ defmodule PhoenixTest.IndexLive do
       </button>
     </form>
 
-    <form id="full-form" phx-submit="save-form">
+    <form id="full-form" phx-change="validate-form" phx-submit="save-form">
       <label for="first_name">First Name</label>
-      <input id="first_name" name="first_name" />
+      <input id="first_name" name="first_name" phx-update="ignore" />
```
ftes, line 132: Live driver doesn't correctly handle "uncontrolled" form inputs (which don't have their `value` attr controlled, e.g. via a `Form`/`Changeset`).

In reality, any phx HTML patch causes their value to be reset.
E.g. checkbox checked state is lost, input text is lost.

Usually not an issue in reality.
But an issue with our test setup when running with real browser.

test/support/index_live.ex
```diff
@@ -286,13 +286,13 @@ defmodule PhoenixTest.IndexLive do
         <label for="complex-movie">Movie <span>*</span></label>
       </fieldset>
 
-      <label for={@uploads.avatar.ref}>Avatar <span>*</span></label>
-      <.live_file_input upload={@uploads.avatar} />
+      <label for={@uploads.complex_avatar.ref}>Avatar <span>*</span></label>
```
ftes, line 289: Duplicate live input causes errors.

test/support/index_live.ex
```diff
@@ -406,11 +406,9 @@ defmodule PhoenixTest.IndexLive do
         </div>
       </fieldset>
 
-      <label for="pet-select">Choose a pet:</label>
-      <select multiple name="pets" id="pet-select">
-        <option phx-click="select-pet" value="dog">Dog</option>
```
ftes, line 411: phx-click on option not supported

test/support/index_live.ex
```diff
@@ -127,26 +127,26 @@ defmodule PhoenixTest.IndexLive do
       </button>
     </form>
 
-    <form id="full-form" phx-submit="save-form">
+    <form id="full-form" phx-change="validate-form" phx-submit="save-form">
```
ftes, line 130: `phx-change` required for live file inputs.

test/support/page_view.ex
```diff
@@ -66,13 +66,13 @@ defmodule PhoenixTest.PageView do
       data-to="/page/delete_record"
       data-csrf="sometoken"
     >
-      Data-method Delete
+      Data_method Delete
```
ftes, line 69: Workaround

Playwright treats exact: false as case insensitive. PhoenixTest remains case sensitive.

ftes:
New Idea: Performant assertions + helpful and consistent error messages in case of failure:

Happy path (test succeeds): Assert using `playwright` primitives.
Test fails: Generate error by calling `PhoenixTest.Assertions` or `PhoenixTest.Element.*.find*` function.

ftes:
Replaced `also_test_js` macro with ExUnit 1.18 `parameterized` tests.
As a side effect, playwright tests only run for the CI test matrix `elixir=main1.18` (1.18) entry. And not for the older versions.

Given this would be a driver in "beta" state anyway, I think that's fine.

image

.github/workflows/ci.yml
```diff
@@ -63,11 +63,20 @@ jobs:
...
 
+    - name: Build assets for browser tests
+      run: mix do assets.setup, assets.build
+
+    - name: Install playwright browsers
+      run: npm exec --prefix priv/static/assets playwright install --with-deps
+
     - name: Run tests
-      run: mix test
+      run: "mix test || if [[ $? = 2 ]]; then mix test --failed; else false; fi"
```
ftes, line 82: Browser (network) based tests tend to be flaky.

Options:
1. Increase timeouts -> test suite takes longer to run (every `refute` has to wait for time out)
2. Keep timeout small, re run failed tests

I opted for 2.

test/phoenix_test/live_test.exs
```diff
@@ -1,42 +1,30 @@
...
-        conn
-        |> visit("/live/index")
-        |> click_button("Change page title")
-        |> PhoenixTest.Driver.render_page_title()
-
-      assert title == "Title changed!"
+      conn
+      |> visit("/live/index")
+      |> click_button("Change page title")
+      |> assert_has("title", text: "Title changed!")
```
ftes, line 19: Using `render_page_title` fails when running via Playwright.
Since the live title patch is not visible immediately in the browser.

Rather, use `assert_has`, which uses a retry mechanism.

.tool-versions
```diff
@@ -1,2 +1,2 @@
 erlang 27.0
-elixir 1.17.2-otp-27
+elixir main
```
ftes, line 2: We need the 1.18 parameterized tests!

lib/phoenix_test.ex
```diff
@@ -230,6 +230,10 @@ defmodule PhoenixTest do
     end
   end
 
+  def visit(driver, path) when is_struct(driver) do
```
ftes, line 233: This seems like the easiest, extensible way of adding custom drivers.

test/phoenix_test/live_test.exs
```diff
@@ -299,7 +304,7 @@ defmodule PhoenixTest.LiveTest do
       |> within("#email-form", fn session ->
         fill_in(session, "Email", with: "someone@example.com")
       end)
-      |> assert_has(input(label: "Email", value: "someone@example.com"))
+      |> assert_has("#form-data", text: "email: someone@example.com")
```
ftes, line 307: Was a bad test:

The `input` is not controlled (`value` not set by server).
When a user inputs a value in the browser, the HTML `value` attribute is not changed.

Fix: Assert on the server-rendered `#form-data` instead.

ftes:
Progress update:

I've slowly been converting Wallaby tests in a production project to PhoenixTest + Playwright.

- I've added some more features along the way (especially: record a Playwright trace for failing tests)
- The foundation of this PR seems to be pretty solid.
- Feedback from more users would be good.

I'm debating how to continue this work.
@germsvel you seem to be pretty tight on time atm.
So I'll keep my work separate for the time being.

@anyone else: If you're interested in using this, let me know.
Then I'll consider publishing the standalone Playwright driver to github or hex.pm.

germsvel:
@ftes love the progress! I'm sorry I haven't had the time to review the PR. Because of the size of it, I need a good few hours to review it properly and give helpful feedback.

Something I've been thinking about: what if we make the protocol layer open for external libraries? We would have to change some of the protocol so it can be fully implemented by external libraries, but I think that would allow us to build other adapters at will. So, for example, you could add this adapter as an external library (e.g. `phoenix_test_playwright` or something like that), and I don't have to be the blocker to all this good work. 

What do you think?

ftes:
> what if we make the protocol layer open for external libraries

I think that's a great idea - as long as it doesn't make PhoenixTest too complicated.

We could start by listing necessary changes (new issue?):
- support `visit/2` for custom drivers (e.g. #149)

lib/phoenix_test/playwright.ex
```diff
@@ -0,0 +1,398 @@
...
+  to run tests concurrently in different browsers.
+
+
+  ## Known limitations and inconsistencies
+
+  - `PhoenixTest.select/4` option `exact_option` is not supported
+  - Playwright driver is less strict than `Live` and `Static` drivers. It does not raise errors
+    - when visiting a page that returns a `404` status
+    - when interactive elements such as forms and buttons are missing essential attributes (`phx-click`, `phx-submit`, `action`)
+  - A few small bugs
```
germsvel, line 45: Do you know what these bugs are? Do we expect to solve them, and therefore is better not to mention it at all?

lib/phoenix_test/playwright.ex
```diff
@@ -0,0 +1,398 @@
...
+      trace_dir: "tmp"
+    ],
+    timeout_ms: 1000
+  ```
+
+  ## Ecto SQL.Sandbox
+
+  `PhoenixTest.Case` automatically takes care of this.
+  It passes a user agent referencing your Ecto repos.
+  This allows for [concurrent browser tests](https://hexdocs.pm/phoenix_ecto/main.html#concurrent-browser-tests).
```
germsvel, line 70: 🥳

lib/phoenix_test/playwright.ex
```diff
@@ -0,0 +1,398 @@
...
+  defdelegate assert_has(session, selector), to: Playwright
+  defdelegate assert_has(session, selector, opts), to: Playwright
+  defdelegate refute_has(session, selector), to: Playwright
+  defdelegate refute_has(session, selector, opts), to: Playwright
+
+  def assert_path(session, path), do: Playwright.retry(fn -> Assertions.assert_path(session, path) end)
+  def assert_path(session, path, opts), do: Playwright.retry(fn -> Assertions.assert_path(session, path, opts) end)
+  def refute_path(session, path), do: Playwright.retry(fn -> Assertions.refute_path(session, path) end)
+  def refute_path(session, path, opts), do: Playwright.retry(fn -> Assertions.refute_path(session, path, opts) end)
+end
```
germsvel, line 398: This is a really cool module. I like it. 🥇

lib/phoenix_test/playwright/browser.ex
```diff
@@ -0,0 +1,20 @@
+defmodule PhoenixTest.Playwright.Browser do
+  @moduledoc """
+  Interact with a Playwright `Browser`.
+
+  There is no official documentation, since this is considered Playwright internal.
```
germsvel, line 5: Since this is private, I tend to make these `@moduledoc false`, and if you want to leave the documentation, leave it at the top as comments. That way, the module isn't visible from hexdocs, and people don't use it directly. 

As much as possible, I try not to expose modules that are private so that we can change private APIs more freely without having to do major version bumps. 

What do you think?

lib/phoenix_test/playwright/browser_context.ex
```diff
@@ -0,0 +1,48 @@
...
+
+  @doc """
+  Start tracing. The results can be retrieved via `stop_tracing/2`.
+  """
+  def start_tracing(context_id, opts \\ []) do
+    opts = Keyword.validate!(opts, screenshots: true, snapshots: true, sources: true)
+    tracing_id = initializer(context_id).tracing.guid
+    post(method: :tracingStart, guid: tracing_id, params: Map.new(opts))
+    post(method: :tracingStartChunk, guid: tracing_id)
+    :ok
```
germsvel, line 29: I don't know anything about tracing in Playwright, but looks interesting!

lib/phoenix_test/playwright/connection.ex
```diff
@@ -0,0 +1,225 @@
...
+
+  defp add_guid_ancestors(state, %{method: "__create__"} = msg) do
+    child = msg.params.guid
+    parent = msg.guid
+    parent_ancestors = Map.get(state.guid_ancestors, parent, [])
+
+    Map.update!(state, :guid_ancestors, &Map.put(&1, child, [parent | parent_ancestors]))
+  end
+
+  defp add_guid_ancestors(state, _), do: state
```
germsvel, line 189: This is just my lack of Playwright knowledge -- what are these ancestors and why do we have to do this?

lib/phoenix_test/playwright/connection.ex
```diff
@@ -0,0 +1,225 @@
...
+    msg = msg |> Enum.into(default) |> update_in(~w(params timeout)a, &(&1 || timeout()))
+    timeout = msg.params.timeout
+    timeout_with_grace_period = timeout + @playwright_timeout_grace_period_ms
+    GenServer.call(@name, {:post, msg}, timeout_with_grace_period)
+  end
+
+  @doc """
+  Get all past received messages for a playwright `guid` (e.g. a `Frame`).
+  The internal map used to track these messages is never cleaned, it will keep on growing.
+  Since we're dealing with (short-lived) tests, that should be fine.
```
germsvel, line 79: Out of curiosity, why do we need to keep track of all of those messages? We don't remove them when they're "consumed" once we send them to the subscribers? (not saying we should, just trying to understand the reason behind the changes)

lib/phoenix_test/playwright/frame.ex
```diff
@@ -0,0 +1,125 @@
+defmodule PhoenixTest.Playwright.Frame do
```
germsvel, line 1: Nice module! 👍

lib/phoenix_test/playwright/message.ex
```diff
@@ -0,0 +1,45 @@
+defmodule PhoenixTest.Playwright.Message do
+  @moduledoc """
+  Parse playwright messages.
+  A single `Port` message can contain multiple Playwright messages and/or a fraction of a message.
+  Such a message fraction is stored in `bufffer` and continued in the next `Port` message.
```
germsvel, line 5: Tiny typo. Thank for the comment! Helpful to know why we need this. 

```suggestion
  Such a message fraction is stored in `buffer` and continued in the next `Port` message.
```

lib/phoenix_test/playwright/port.ex
```diff
@@ -0,0 +1,77 @@
...
+      To resolve this please
+      1. Install playwright, e.g. `npm i playwright`
+      2. Configure the path correctly, e.g. in `config/text.exs`: `config :phoenix_test, playwright: [cli: "assets/node_modules/playwright/cli.js"]`
+      """
+
+      raise ArgumentError, msg
+    end
+
+    cmd = "run-driver"
+    port = Port.open({:spawn, "#{cli} #{cmd}"}, [:binary])
```
germsvel, line 31: I asked this before in the PR, but worth asking here again: have you run into any zombie OS processes? 

I know the [Port docs](https://hexdocs.pm/elixir/Port.html#module-zombie-operating-system-processes) mention a fix around it. I've used it in the past, and it worked quite well. Not sure if we need it here, but just pointing it out in case it's helpful.

lib/phoenix_test/playwright/port.ex
```diff
@@ -0,0 +1,77 @@
...
+
+  defp serialize(message) do
+    Jason.encode!(message)
+  end
+
+  defp atom_keys(map) when is_map(map) do
+    Map.new(map, fn
+      {k, v} when is_map(v) -> {String.to_atom(k), atom_keys(v)}
+      {k, list} when is_list(list) -> {String.to_atom(k), Enum.map(list, fn v -> atom_keys(v) end)}
+      {k, v} -> {String.to_atom(k), v}
```
germsvel, line 72: Any concerns here with using `String.to_atom/1`? It's typically not a good idea with data we don't control, right? Are we at risk of hitting the max number of atoms at some point? And could we use `to_exiting_atom/1` instead? Or do we do some other type of mapping?

lib/phoenix_test/playwright/selector.ex
```diff
@@ -0,0 +1,42 @@
...
+  - https://playwright.dev/docs/other-locators
+  - https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/client/locator.ts
+  - https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/utils/isomorphic/locatorUtils.ts
+  """
+
+  def concat(left, :none), do: left
+  def concat(left, right), do: "#{left} >> #{right}"
+
+  def unquote(:and)(left, :none), do: left
+  def unquote(:and)(left, right), do: concat(left, "internal:and=#{Jason.encode!(right)}")
```
germsvel, line 22: Is there a different function name we could use here so it doesn't conflict with `and`, and so we don't have to `unquote` here? Or do you think it's worth it so we can have the 1-1 mapping with Playwright? (I think that's why we're doing it, right?)

germsvel (commented):
@ftes I haven't finished reviewing the full PR (I haven't looked at the test changes), but hopefully this is a good start (regardless of whether this comes directly into PhoenixTest or we make it a separate library). 

Seeing how advanced this is, I'm open to including it directly into PhoenixTest (if you want to do that). I have an alternative solution in progress using Chrome Driver Protocol directly, but in that solution, we still have to provide the path to Chromium. So, if we're going to provide the path to a binary/cli thing, then maybe it makes more sense to just go full on with playwright. 

If you prefer to keep it separate (which I also totally understand. Might be easier to move faster), I'd be happy for us to work more towards making the Driver protocol open for that. 

One thing that I'm not super clear on yet is -- how close to feature parity is this? Do you consider it ready to merge ASAP, and we're good to go? Or are there things you'd like to finish before this could become part of PhoenixTest? 

Any and all thoughts welcome!

config/test.exs
```diff
@@ -1,6 +1,14 @@
 import Config
 
-config :phoenix_test, :endpoint, PhoenixTest.Endpoint
+config :phoenix_test,
+  endpoint: PhoenixTest.Endpoint,
+  ecto_repos: [PhoenixTest.Repo],
+  otp_app: :phoenix_test,
+  playwright: [
+    browser: [browser: :chromium, headless: System.get_env("PLAYWRIGHT_HEADLESS", "t") in ~w(t true)],
+    cli: "priv/static/assets/node_modules/playwright/cli.js",
```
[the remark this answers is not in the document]
ftes: Good question. I honestly have no idea how stable the playwright protocol is.
Since its considered internal, there probably are no strong guarantees, so its difficult to say.

lib/phoenix_test/case.ex
```diff
@@ -0,0 +1,120 @@
...
+  setup_all context do
+    trace = Application.fetch_env!(:phoenix_test, :playwright)[:trace]
+
+    case context do
+      %{playwright: true} ->
+        [browser_id: Case.Playwright.launch_browser(@playwright_opts), trace: trace]
+
+      %{playwright: opts} when is_list(opts) ->
+        opts = Keyword.merge(@playwright_opts, opts)
+        [browser_id: Case.Playwright.launch_browser(opts), trace: trace]
```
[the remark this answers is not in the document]
ftes: Yes, exactly. The browser is only launched once.
And then one session within that browser per test (think new user profile + window).

And `Connection.ensure_started` provides lazy launching of the `playwright` node.js server, so it is only started if actually required by a test.

lib/phoenix_test/playwright/port.ex
ftes: I haven't seen any yet.
In general my experience with playwright has been far better than with Wallaby.
The playwright server seems to take better care of terminating the browser if a test is cancelled mid-run.

I'll keep my eyes open.

lib/phoenix_test/playwright/port.ex
ftes: Good question.
Since we're in a context of tests, I wouldn't worry too much.
We're not talking about a long-running system.

I think the keys we're converting to atoms are all playwright defined, i.e. there's a limited well-defined amount.
But since I don't know all playwright message types I can't say for sure whether or not e.g. some "user-defined" HTML node attribute might not pop up as a nested message key somewhere.

But even then, we're talking about tests and "user provided" would (mostly)  mean "provided by the test set up".

I wouldn't want to use `String.to_existing_atom` because I'm worried that users might get `ArgumentErrors` if they trigger playwright messages with new keys (e.g. because they're using `unwrap` in ways that triggers new message types).

The safe way would be to keep everything as strings, but that would just make the code so much more verbose.

lib/phoenix_test/playwright/selector.ex
ftes: Yes, I tried to keep the `Selector`, `Frame` modules as minimal playwright wrappers which ideally reflect playwright concepts 1:1.

We could always go with `and_` or something in that direction.

ftes:
> how close to feature parity is this

Good question! I've been using a copy of this for some weeks in a client project now.
A few small things have changed after migrating dozens of Wallaby tests.

What I've gotten rid of there is the calls to `Assertions`. 
In this PR I tried to keep the exact same interface - including generating the exact same errors.
While this provides the best user experience, it also introduces more coupling that might make it more difficult to change code in future (no matter if the playwright driver is internal or external).

So the driver is definitely usable.
But I expect changes as soon as further people use it.

I'm undecided:
- Pro internal: Running the existing test suite against the Playwright driver is great to ensure consistent behaviour
- Con internal: As mentioned above, I'm unsure whether trying to mimic`Live`/`Static` drivers 100% is a good goal. In some places (e.g. role-based selectors), it is easiest and most efficient to use existing Playwright concepts. Trying to be 100% compatible precludes that option. Keeping things separate makes that dividing line more apparent.

Taking everything into consideration, I think it might be best to keep it separate.
And ensure that `PhoenixTest` is external-driver-friendly.

Excited for the vanilla CDP driver! It will be great to compare and see the different tradeoffs.

ftes:
Closing this for now.

PR #146, opened 2024-11-05
Live redirect copy headers

joeytrapp:
The added behavior of copying headers wasn't working in my application, and after toying around with a local path dependency, I found that in the `:live_redirect` case the headers weren't being copied. While adding the call to `recycle/2` with the list of all headers made things work again, I wasn't sure why all of your tests were passing.

Included in this PR is a scenario the more closely mirrors what is happening in my application, where the cookie session is modified in a Plug. This scenario provokes the error, and my second commit adds the additional header copy to pass the tests. I suspect the test case could be integrated into the test suite better.

### DIFF
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -328,9 +328,10 @@ defmodule PhoenixTest.Live do
 
   defp maybe_redirect({:error, {:live_redirect, %{to: path}}} = result, session) do
     session = %{session | current_path: path}
+    conn = session.conn
 
     result
-    |> follow_redirect(session.conn)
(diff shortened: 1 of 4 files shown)


joeytrapp:
@germsvel is there more detail I can provide here?

germsvel (approved):
This looks good. Thanks @joeytrapp! 🙌

joeytrapp:
I'll have a look at these new failures after the rebase.

germsvel:
@joeytrapp no worries. This wasn't on you. I merged too many PRs in one shot and something conflicted. I'll fix this.

PR #148, opened 2024-11-18
Document click_button/link exact=false semantics

ftes:
Use same exact-match semantics for `click_link` and `click_button` for `Static` and `Live` drivers.

## Live
The `Live` driver has `exact=false` semantics for `click_link` and `click_button` because it uses [`Phoenix.LiveViewTest.element/3`](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveViewTest.html#element/3).

> If the text filter is a string or a regex, it will match any element that contains the string (including as a substring) or matches the regex.
> https://hexdocs.pm/phoenix_live_view/Phoenix.LiveViewTest.html#element/3

## Static
Fix: Align with live driver by forcing `exact=false`.

### DIFF
--- a/lib/phoenix_test/static.ex
+++ b/lib/phoenix_test/static.ex
@@ -59,7 +59,7 @@ defmodule PhoenixTest.Static do
     link =
       session
       |> render_html()
-      |> Link.find!(selector, text)
+      |> Link.find!(selector, text, exacxt: false)
 
     if Link.has_data_method?(link) do
       form =
@@ -81,7 +81,7 @@ defmodule PhoenixTest.Static do
     active_form = session.active_form
 
     html = render_html(session)
-    button = Button.find!(html, selector, text)
+    button = Button.find!(html, selector, text, exact: false)
 
     if Button.has_data_method?(button) do
       form =
--- a/lib/phoenix_test/static.ex
+++ b/lib/phoenix_test/static.ex
@@ -59,7 +59,7 @@ defmodule PhoenixTest.Static do
     link =
       session
       |> render_html()
-      |> Link.find!(selector, text, exacxt: false)
+      |> Link.find!(selector, text, exact: false)
 
     if Link.has_data_method?(link) do
       form =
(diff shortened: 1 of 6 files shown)


test/phoenix_test/static_test.exs
```diff
@@ -73,6 +73,13 @@ defmodule PhoenixTest.StaticTest do
       |> assert_has("h1", text: "Main page")
     end
 
+    test "fins by substring", %{conn: conn} do
```
germsvel, line 76: Tiny typo: 

```suggestion
    test "finds by substring", %{conn: conn} do
```

lib/phoenix_test/static.ex
```diff
@@ -59,7 +59,7 @@ defmodule PhoenixTest.Static do
     link =
       session
       |> render_html()
-      |> Link.find!(selector, text)
+      |> Link.find!(selector, text, exacxt: false)
```
germsvel, line 62: This typo should make it so the test fails since it should be matching by `exact: true`. But I think the test is passing because the default in `Query.find/3` is `false`. 

I think we might have different semantics in different places. I think this `Link.find!` is eventually  using https://github.com/germsvel/phoenix_test/blob/ddacde8bff7be17c4b58782a3c9bb3da26ea50c9/lib/phoenix_test/query.ex#L518

But places that use `find_by_label!` seem to be defaulting to `exact: true` (https://github.com/germsvel/phoenix_test/blob/ddacde8bff7be17c4b58782a3c9bb3da26ea50c9/lib/phoenix_test/query.ex#L159)

So... I think the core of your PR might be true, but I'm not sure this is fixing it?

germsvel (commented):
Thanks for opening this PR. I agree that the semantics should be the same. 

But I'm not sure if the PR is fully fixing it. I left a comment where there's a typo on `exact: false` which makes it seem like one of the tests you added should fail. But it's not failing, and I think it's not because there seem to be two places with defaults on in `Query.find!` (eventually the filtering of text has the default) and one in `Query.find_by_label!`.

lib/phoenix_test/static.ex
ftes: Interesting. The only explanation I have is that I found the inconsistency in a release version and it has since been fixed ("accidentally")?

I've reverted actual code changes and changed the PR to just document the status quo.

Good catch!

PR #149, opened 2024-11-18
Support visit/2 for custom drivers

ftes:
In addition to the auto detection of `Live` vs. `Static` via `Plug.Conn`.

Usage (add a test?):

```ex
setup_args
|> MyDriver.build()
|> visit("/path")
...
```

## Alternative (that wouldn't require changes to PhoenixTest)

Instead of adding a `visit/2` to the Driver protocol, custom drivers could implement their own variant.
No real downside. Apart from making it more or less obvious that custom drivers are supported. 

```ex
setup_args
|> MyDriver.visit("/path")
...
```

### DIFF
--- a/lib/phoenix_test.ex
+++ b/lib/phoenix_test.ex
@@ -213,7 +213,7 @@ defmodule PhoenixTest do
   LiveView or a static view. You don't need to worry about which type of page
   you're visiting.
   """
-  def visit(conn, path) do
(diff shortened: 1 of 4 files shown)


germsvel:
Thanks for doing this! I opted for #160 to avoid the need to `raise`. It's _very minor_ improvement, but I think I like it better since it at least gives us a clue as to what `Live` and `Static` do with `visit`.

PR #151, opened 2024-12-03
Move default link/button selectors into Driver

ftes:
Let Driver choose default selector for `click_link/2` and `click_button/2`.

Allow custom drivers such as Playwright to apply their own default selectors.
Don't require custom drivers to know PhoenixTest internal `locator`s


ftes:
The same should be done for `check/2` etc.
Which would allow other drivers (playwright) to use better default locators (`getByRole('checkbox')`).

lib/phoenix_test/driver.ex
```diff
@@ -2,7 +2,9 @@ defprotocol PhoenixTest.Driver do
   @moduledoc false
   def render_page_title(session)
   def render_html(session)
+  def click_link(session, text)
   def click_link(session, selector, text)
+  def click_button(session, text)
```
germsvel, line 7: I'm curious what you think about _not_ exposing the 2-arity versions of `click_link` and `click_button`. 

My thought is to try to keep the API as simple as possible for libraries that want to implement it. They should know that they'll _always_ get a `selector`. 

PhoenixTest just has a nice sugar version of `click_link/2` and `click_button/2` that passes the `"a"` and button's selector. I realize the button selector is slightly more complicated (i.e. not just `"button"`), but maybe that's something that the external drivers benefit from PhoenixTest?

germsvel (commented):
Thanks for doing this. I have a minor question. Curious what you think.

lib/phoenix_test/driver.ex
ftes: I see two problems with the defaults applied at `PhoenixTest` library layer
1. It prevents external drivers (Playwright) from using better default selectors: `internal:role=link`
2. It leaks PhoenixTest internals like `%Locators.Button{}` into external drivers (right now there's no need to do that, although a point can be made that at some point every driver should understand the PhoenixTest concept of locators)

germsvel: Yeah, those are good points. I like the idea of the external drivers being able to use their own role-based selectors and stuff like that. Worth the trade-off here. 👍

germsvel:
Thanks for this work!

PR #155, opened 2024-12-05
Fix: phx-trigger-action patch test

ftes:
Broken by #153 

Sorry about that.


germsvel (approved):
Thanks for the fix. No worries. I should've caught it too. Trying to merge too many PRs in one shot 😄

PR #156, opened 2024-12-05
Move default input selectors into Driver

ftes:
Like #151, but for the input action functions (`fill_in`, `select` etc.).


germsvel (approved):

germsvel:
Thanks @ftes!

PR #158, opened 2024-12-09
Normalize whitespace (find label with line breaks)

ftes:
Similar to other feature testing libraries.

Wallaby
>  Matches any element by its inner text.
`~s{.//*[contains(normalize-space(text()), "#{selector}")]}`
[xpath.ex](https://github.com/elixir-wallaby/wallaby/blob/c144c92e8bc1812c1f712cc1f1c652c4d2c7ce4b/lib/wallaby/query/xpath.ex#L80)
https://developer.mozilla.org/en-US/docs/Web/XPath/Functions/normalize-space

Playwright
> Matching by text always normalizes whitespace, even with exact match. For example, it turns multiple spaces into one, turns line breaks into spaces and ignores leading and trailing whitespace.
https://playwright.dev/docs/locators#get-by-text

### DIFF
--- a/lib/phoenix_test/html.ex
+++ b/lib/phoenix_test/html.ex
@@ -5,7 +5,9 @@ defmodule PhoenixTest.Html do
     Floki.parse_document!(html)
   end
 
-  def text(element), do: element |> Floki.text() |> String.trim()
+  def text(element) do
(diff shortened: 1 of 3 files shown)


germsvel:
Thanks @ftes!

PR #162, opened 2024-12-31
feat: :sparkles: trigger phx-change on upload

### LINKED ISSUE #150: Select a file without uploading it
`upload/4` does not behave as I would expect. I wrote a test to check if a selected file in a list of files I want to upload is displayed with its name and file size. 
However, the upload itself should not be performed yet. The description of `upload/4` says that it can be followed by a `click_button/3` or `submit/1`. From this I concluded that the upload has not yet been started and `upload/4` is used to ‘preselect’ the files. I had also discovered that `Phoenix.LiveViewTest.render_upload/3` is responsible for this, as this is called with `upload/4`.

Are my expectations wrong or am I just using `upload/4` incorrectly?

Otherwise, many thanks for this super cool library. Great work!

dweremeichik:
When a file is selected in the browser, phx-change is triggered on the form. This changes the behaviour of upload/4 to be closer to the browsers behaviour. 

BREAKING CHANGE: users who previously worked around this issue may have phx-change called twice.

### DIFF
--- a/lib/phoenix_test.ex
+++ b/lib/phoenix_test.ex
@@ -1095,9 +1095,12 @@ defmodule PhoenixTest do
   @doc """
   Upload a file.
 
-  If the form is a LiveView form, this will perform a live file upload.
+  If the form is a LiveView form, this will perform a live file upload and trigger the associated `phx-change` event.
 
   This can be followed by a `click_button/3` or `submit/1` to submit the form.
+
+  > LiveView Note: If `allow_upload/3` has a `:progress` callback that issues a navigate or redirect,
+  > it will be used instead of any navigate, redirect or patch from the upload's `phx-change` event.
 
   ## Options
 
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -259,11 +259,24 @@ defmodule PhoenixTest.Live do
       type: mime_type
     }
 
+    upload_progress_result =
+      session.view
+      |> file_input(form.selector, live_upload_name, [entry])
+      |> render_upload(file_name)
+      |> maybe_throw_upload_errors(session, file_name, live_upload_name)
+
     session.view
-    |> file_input(form.selector, live_upload_name, [entry])
-    |> render_upload(file_name)
-    |> maybe_throw_upload_errors(session, file_name, live_upload_name)
-    |> maybe_redirect(session)
+    |> form(form.selector)
+    |> render_change(%{"_target" => field.name})
+    |> progress_redirect_or_change_result(upload_progress_result, session)
+  end
+
+  defp progress_redirect_or_change_result(_change_result, {:error, _} = upload_progress_result, session) do
+    maybe_redirect(upload_progress_result, session)
+  end
+
+  defp progress_redirect_or_change_result(change_result, _upload_progress_result, session) do
+    maybe_redirect(change_result, session)
   end
 
   defp maybe_throw_upload_errors({:error, [[_id, error]]}, session, file_name, live_upload_name) do
--- a/test/support/web_app/index_live.ex
+++ b/test/support/web_app/index_live.ex
@@ -138,7 +138,7 @@ defmodule PhoenixTest.WebApp.IndexLive do
       </button>
     </form>
 
-    <form id="full-form" phx-submit="save-form">
+    <form id="full-form" phx-submit="save-form" phx-change="upload-change">
       <label for="first_name">First Name</label>
       <input id="first_name" name="first_name" />
 
@@ -286,7 +286,7 @@ defmodule PhoenixTest.WebApp.IndexLive do
       <input id="email-on-change" name="email" />
     </form>
 
-    <form id="complex-labels" phx-change="save-form" phx-submit="save-form">
+    <form id="complex-labels" phx-change="change-form" phx-submit="save-form">
       <label for="complex-name">
         Name <span>*</span>
       </label>
@@ -320,7 +320,7 @@ defmodule PhoenixTest.WebApp.IndexLive do
       <button type="submit">Save</button>
     </form>
 
-    <form id="same-labels" phx-submit="save-form" phx-change="save-form">
+    <form id="same-labels" phx-submit="save-form" phx-change="change-form">
       <fieldset name="like-elixir">
         <legend>Do you like Elixir:</legend>
 
@@ -511,11 +511,6 @@ defmodule PhoenixTest.WebApp.IndexLive do
       <button phx-click="navigate-and-trigger-form">Navigate and trigger action</button>
     </form>
 
-    <form id="tiny-upload-form">
-      <label for={@uploads.tiny.ref}>Tiny</label>
-      <.live_file_input upload={@uploads.tiny} />
-    </form>
-
     <form id="conditional-inputs" phx-change="save-form" phx-submit="save-form">
       <label>
         <input name="hide_to_remove" type="checkbox" value="on" /> Hide to remove
@@ -526,6 +521,25 @@ defmodule PhoenixTest.WebApp.IndexLive do
       <label>
         To keep <input name="to_keep" type="text" />
       </label>
+    </form>
+
+    <form id="tiny-upload-form">
+      <label for={@uploads.tiny.ref}>Tiny</label>
+      <.live_file_input upload={@uploads.tiny} />
+    </form>
+
+    <form id="upload-change-form" phx-change="upload-change">
+      <label for={@uploads.avatar.ref}>Avatar</label>
+      <.live_file_input upload={@uploads.avatar} />
+    </form>
+
+    <div :if={@upload_change_triggered} id="upload-change-result">
+      phx-change triggered on file selection
+    </div>
+
+    <form id="upload-redirect-form" phx-change="upload-change">
+      <label for={@uploads.redirect_avatar.ref}>Redirect Avatar</label>
+      <.live_file_input upload={@uploads.redirect_avatar} />
     </form>
     """
   end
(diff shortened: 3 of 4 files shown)


dweremeichik:
It's probably worth mentioning that if this library decides to implement the optional upload percentage (argument 3) of `Phoenix.LiveViewTest.render_upload/3` in the future, `phx-change` could be called multiple times which would be undesirable.

germsvel:
@dweremeichik thanks for taking a stab at this! 🙌

 I haven't looked at it yet, but will do so as soon as I can. I want to go test a bunch of uploads -- I haven't had to deal with them in a little while, so I'm unclear as to the pain points and what would be great ergonomics. 

In the meantime, I wanted to share some thoughts on what you said: 

> It's probably worth mentioning that if this library decides to implement the optional upload percentage (argument 3) of Phoenix.LiveViewTest.render_upload/3 in the future, phx-change could be called multiple times which would be undesirable.

Do you think that in the case we could pattern match on the percent passed, and we could trigger the `phx-change` only if the user passes `100`? 

Not saying it has to be like this, but curious if that would solve the issue you're thinking about: 

```elixir
session
|> upload("Avatar", "/path/to/file", percent: 50) # <= doesn't trigger phx-change

session
|> upload("Avatar", "/path/to/file", percent: 100) # <= triggers phx-change
```

That is, of course, so long as that's what Phoenix does behind the scenes.

germsvel (commented):
Thanks so much for opening this PR! 🙌 It's looking good. 

I just have a question for something I don't quite understand in the changes in the comments below.

test/support/web_app/index_live.ex
```diff
@@ -278,7 +278,7 @@ defmodule PhoenixTest.WebApp.IndexLive do
       <input id="email-on-change" name="email" />
     </form>
 
-    <form id="complex-labels" phx-change="save-form" phx-submit="save-form">
+    <form id="complex-labels" phx-change="change-form" phx-submit="save-form">
```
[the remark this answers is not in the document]
dweremeichik: `phx-change` never used to get called for any of the `upload/3` tests that used `"save-form"`. Now that it does, it broke those tests.

The `handle_event/3` callback for `"save-form"` calls  `consume_uploaded_entries/3` [here](https://github.com/germsvel/phoenix_test/pull/162/files/318db75e7b3d1e22c6a02a822972f13c52330d68#diff-5adad04c8b54890cfd27c70faa03249bc8617cb3bb3f7c3d9886b28c853c494cL550-L561) . When `phx-submit` and `phx-change` are both set to `"save-form"`, `consume_uploaded_entries/3` gets called twice which will cause errors. I split it into two separate events to call out that the events should have different behaviours. However `"change-form"` is really just a copy of the last 6 lines of `"save-form"`; which previous tests had access to.

The two code comments below show why this wasn't a problem previously:

```elixir
test "can target a label with exact: false", %{conn: conn} do
      conn
      |> visit("/live/index")
      |> within("#complex-labels", fn session ->
        session
        ##### upload/3 never used to trigger `phx-change`, but does now #####
        |> upload("Avatar", "test/files/elixir.jpg", exact: false) 
        ##### click_button/1 has always triggered `phx-submit` #####
        |> click_button("Save")
      end)
      |> assert_has("#form-data", text: "avatar: elixir.jpg")
    end
```

lib/phoenix_test/live.ex
```diff
@@ -263,10 +263,22 @@ defmodule PhoenixTest.Live do
...
-    session.view
+    upload_progress_result = session.view
     |> file_input(form.selector, live_upload_name, [entry])
     |> render_upload(file_name)
-    |> maybe_redirect(session)
+
+    session.view
+    |> form(form.selector)
+    |> render_change(%{"_target" => field.name})
+    |> progress_redirect_or_change_result(upload_progress_result, session)
```
dweremeichik, line 273: For clarity, this line chooses whether the result of `render_upload` or `render_change` is used. The result from `render_upload` is only currently used if it instructs a redirect.

dweremeichik:
> In the meantime, I wanted to share some thoughts on what you said:
> 
> > It's probably worth mentioning that if this library decides to implement the optional upload percentage (argument 3) of Phoenix.LiveViewTest.render_upload/3 in the future, phx-change could be called multiple times which would be undesirable.
> 
> Do you think that in the case we could pattern match on the percent passed, and we could trigger the `phx-change` only if the user passes `100`?

I believe `phx-change` should be called at a similar time to when it would be triggered by a file input in the browser, which would be on the initial call to `upload`. We could probably check the upload assigns to see if an upload already exists. If it does, then we could update the percentage to whatever the user passes.

test/support/web_app/index_live.ex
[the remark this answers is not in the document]
germsvel: Got it. That makes sense. Thanks for the explanation!

germsvel:
> I believe phx-change should be called at a similar time to when it would be triggered by a file input in the browser

Totally agree. We should make the test as close as to the "real" implementation as possible. That way the tests don't mislead us. 

> We could probably check the upload assigns to see if an upload already exists. If it does, then we could update the percentage to whatever the user passes.

👍 I imagine we can leave that for a follow up if people are interested in that?

lib/phoenix_test.ex
```diff
@@ -1006,10 +1006,13 @@ defmodule PhoenixTest do
   @doc """
   Upload a file.
 
-  If the form is a LiveView form, this will perform a live file upload.
+  If the form is a LiveView form, this will perform a live file upload and trigger the associated `phx-change` event.
```
germsvel, line 1009: 🙏 thanks for adding to the docs by the way! Much appreciated.

lib/phoenix_test.ex
```diff
@@ -1006,10 +1006,13 @@ defmodule PhoenixTest do
...
 
   This can be followed by a `click_button/3` or `submit/1` to submit the form.
 
+  > LiveView Note: If `allow_upload/3` has a `:progress` callback that issues a navigate or redirect,
+  > it will be used instead of any navigate, redirect or patch from the upload's `phx-change` event.
```
germsvel, line 1014: Is this the behavior in LiveViews right now? A navigation from a `:process` callback takes precedence over what happens in the `phx-change` handle-event callback?

germsvel (commented):
One more question (just for my own understanding) but otherwise I think this is good to go! 

(I added a commit that ran `mix format` to fix the build)

lib/phoenix_test.ex
dweremeichik: It is not. It is an unfortunate side effect of `render_upload`. In writing my response to this, I realized that I ended up over engineering this due to the broken tests that I had at the time and not paying full attention to the whole [view immutability behaviour](https://www.germanvelasco.com/blog/is-the-test-liveview-immutable-or-not) (I originally read about it elsewhere, but thank you for documenting it!). I didn't realize it, but it appears that live view will not overwrite existing redirect requests, the first redirect requested seems to win. There is unnecessary code behind this note that I will refactor and remove.

I think this can be made nicer by refactoring `upload` and changing the api to accept a percentage from the user. This will allow the user to manipulate when `:progress` events are triggered (as LiveViewTest does), skirting around the issue.

germsvel: >  I didn't realize it, but it appears that live view will not overwrite existing redirect requests, the first redirect requested seems to win.

Makes sense. Do you know what that means in practice here for precedence? 

> There is unnecessary code behind this note that I will refactor and remove.

🙏  thanks! Let me know when you remove that so I can give this another look. 

> I think this can be made nicer by refactoring upload and changing the api to accept a percentage from the user. This will allow the user to manipulate when :progress events are triggered (as LiveViewTest does), skirting around the issue.

Do you think we should head that direction instead of what we're doing here? I'm not sure I understand the issue fully yet to know the difficulties we're encountering. When does LiveView trigger those progress events? Could we emulate that?

dweremeichik: > Do you know what that means in practice here for precedence?

We don't have a choice, `render_upload` will always step on `render_change`, but...

> Do you think we should head that direction instead of what we're doing here? ... When does LiveView trigger those progress events? Could we emulate that?

if we emulate actual browser behaviour I think we won't have to deal with the possible collision. Below I've broken down the two actions that an application end user takes to upload a file. [phx] is the event phoenix triggers because of each respective action. [phx-test] is what we will need to call to get an emulated behaviour in this library. [phx-test-user] is the action a user of this library would take to emulate the end user's actions. 

1. Select file via file input
[phx]: call `phx-change` (possibly start uploading if :auto_upload is true)
[phx-test]: call `render_upload` with a 0 percentage, then `render_change`
[phx-test-user]: call `upload` as normal
1. Click upload (submit) button
[phx]: call `phx-submit`
[phx-test]: call `render_upload` with a 100 percentage
[phx-test-user]: call `click_button` on the upload button or `submit`

of course in between those two actions, we could still allow the user to set whatever percentage they would like. I think this behaviour is okay, because the live view [docs say](https://hexdocs.pm/phoenix_live_view/uploads.html#consume-uploaded-entries:~:text=When%20the%20end%2Duser%20submits%20a%20form%20containing%20a%20live_file_input/1%2C%20the%20JavaScript%20client%20first%20uploads%20the%20file(s)%20before%20invoking%20the%20callback%20for%20the%20form%27s%20phx%2Dsubmit%20event.):

>When the end-user submits a form containing a [live_file_input/1](https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#live_file_input/1), the JavaScript client first uploads the file(s) before invoking the callback for the form's phx-submit event.

germsvel: Do we still need this comment? I wasn't sure if it was supposed to be removed once you removed the unnecessary code you first mentioned.

germsvel (commented):
@dweremeichik do you know what's the status of this PR? 

The changes look good. I'm sorry it took me a while to get back to it -- I think I was assuming something else was needed, but I don't think so? (unless that comment needs to be removed). 

I realize the branch now needs to resolve conflicts. If you don't have time to do that/don't want to do that, let me know, and I can try to push this through the finish line. But if you can, that'd be awesome too.

germsvel:
@dweremeichik I fixed the merge conflicts, but everything else looked good. Thanks for this work! 👍

PR #163, opened 2025-01-01
Allow wildcards in assert_path

larshei:
## Why

To cover scenarios where a user interaction creates a temporary entity which is referred to in the path.

## How

Allow wildcards in `assert_path`.

## What

Example:
```elixir
conn
|> visit("/")
|> click("new chat room")
|> assert_path("/chat/*")
```

Each wildcard `*` maps to exactly one part in the uri, so

`/user/*/profile` will accept `/user/123/profile`, but not `/user/profile` or `/user/123/public/profile`

### DIFF
--- a/lib/phoenix_test.ex
+++ b/lib/phoenix_test.ex
@@ -1322,11 +1322,11 @@ defmodule PhoenixTest do
   |> visit("/users")
   |> assert_path("/users", query_params: %{name: "frodo"})
 
-  # assert we're at a path with a wildcard
+  # assert we're at a path with a wildcard. The wildcard represents a single part of the path.
   conn
-  |> visit("/")
-  |> click("new chat room")
-  |> assert_path("/chat/*")
+  |> visit("/users")
+  |> click("Any User")
+  |> assert_path("/users/*/profile")
   ```
   """
   defdelegate assert_path(session, path), to: Driver
--- a/lib/phoenix_test.ex
+++ b/lib/phoenix_test.ex
@@ -1321,6 +1321,12 @@ defmodule PhoenixTest do
   conn
   |> visit("/users")
   |> assert_path("/users", query_params: %{name: "frodo"})
+
+  # assert we're at a path with a wildcard
+  conn
+  |> visit("/")
+  |> click("new chat room")
+  |> assert_path("/chat/*")
   ```
   """
   defdelegate assert_path(session, path), to: Driver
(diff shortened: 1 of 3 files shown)


germsvel (commented):
Thanks for opening this PR! 

Allowing for wildcards in the path for `assert_path` is an interesting idea. I left a couple of comments in the PR, and I have a larger overarching one. 👇

Do you mind sharing what's your thought process for using a `*` for a wildcard here? Do you know of any prior art in testing libraries? 

When I see `*` in a path, my mind immediately goes to unix path expansion. So, I'm slightly worried people will think it'll match the rest of the path (not just on the next segment). 

I do like the idea that we could extend this at some point to do something like `**/*` to match the rest of the path (or something like that). But I also don't know if that's overkill. 

And then, I also think of some potential alternatives: maybe we could use `:user_id` (Phoenix route-like definitions) or `:_` (similar to Erlang's match specs). I'm not sure I like either of those better than `*`, but just throwing out my thoughts. 

That's why I'm curious, how did you land on `*`?

larshei:
Thanks for the feedback :)

> Do you mind sharing what's your thought process for using a * for a wildcard here? 

Using an asterisk just felt natural to me.
I have not questioned the choice.

RabbitMQ uses * as a wildcard for matching routing instructions:
`my.routing.key` could be matched by `my.*.key` but not by `my.*`. 
They use `#` for matching zero to n segments, which honestly threw me off when I first encountered it.

> I do like the idea that we could extend this at some point to do something like **/* to match the rest of the path

This crossed my mind - with `**` as the pattern as well - but when I was thinking about the implementation, I started to question wether there are use cases for this - The routes should mostly be static in general, with some parts being variables.

But I havent encountered a variable length path. I believe writing `/some/**/path` is probably not worse than `/some/*/long/*/path`. When reading the code, I would also be more happy about latter option, cause I can easily match it to a route found in the Router module.

germsvel:
> Using an asterisk just felt natural to me.

👍 I like that. Natural feeling (i.e. if that's what you'd expect from it) lines up with the principle of least surprise -- so I like that it felt natural.

> RabbitMQ uses * as a wildcard for matching routing instructions:
my.routing.key could be matched by my.*.key but not by my.*.

Nice. I like that prior art. 

> This crossed my mind - with ** as the pattern as well - but when I was thinking about the implementation, I started to question wether there are use cases for this

Yes, I agree. Seems overkill for something we may never need. I like doing the simpler path first. 

>  believe writing /some/**/path is probably not worse than /some/*/long/*/path. When reading the code, I would also be more happy about latter option, cause I can easily match it to a route found in the Router module.

And that's an excellent point. I prefer the latter one too because it's more explicit, and it seems easier to read when looking at what the test is doing. 

Thanks for sharing all the context! Looks good to me. 👍

germsvel (approved):
🙌 thanks for this work!

PR #169, opened 2025-01-15
fix: :bug: pass non-redirect error tuples through `maybe_redirect`

### LINKED ISSUE #167: Uploading an unsupported file type throws a `FunctionClauseError` error in `maybe_redirect/2`
When intentionally uploading a file type that is not in the `:accept` list of `allow_upload/3`. A `FunctionClauseError` gets thrown in `maybe_redirect/2` because it only supports errors matching the redirect pattern.

```sh
 The following arguments were given to PhoenixTest.Live.maybe_redirect/2:

         # 1
         {:error, [["1314", :not_accepted]]}

         # 2
         %PhoenixTest.Live{view: ...

...

Attempted function clauses (showing 4 out of 4):

         defp maybe_redirect({:error, {:redirect, %{to: path}}}, session)
         defp maybe_redirect({:error, {:live_redirect, %{to: path}}} = result, session)
         defp maybe_redirect({:ok, view, _}, session)
         defp maybe_redirect(html, session) when is_binary(html)

...

```

I'll open a PR to add a new clause to allow unknown errors to pass through.

dweremeichik:
`render_upload` can emit errors that are not redirects. This changes `maybe_redirect` and `within` to allow passage of non-redirect errors to the end user.

Fixes #167

### DIFF
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -112,10 +112,15 @@ defmodule PhoenixTest.Live do
   end
 
   def within(session, selector, fun) when is_function(fun, 1) do
-    session
-    |> Map.put(:within, selector)
-    |> fun.()
-    |> Map.put(:within, :none)
+    result =
+      session
+      |> Map.put(:within, selector)
+      |> fun.()
+
+    case result do
+      %{} -> Map.put(result, :within, :none)
+      {:error, _} -> result
+    end
   end
 
   def fill_in(session, label, opts) do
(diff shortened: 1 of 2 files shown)


dweremeichik:
> For example, they can't grab the error and turn around and do something with it.

Only assert against it if they wish.

> What do you think? Is that something we could do here?

I don't see why not. If someone wanted to, they could `assert_raise` the error. Would you see this being an `ArgumentError` or should it be more specific like [`File.Error`](https://hexdocs.pm/elixir/main/File.Error.html)

dweremeichik:
Closing in favour of #177

PR #177, opened 2025-01-27
fix: :bug: throw nice error messages for `render_upload` errors

### LINKED ISSUE #167: Uploading an unsupported file type throws a `FunctionClauseError` error in `maybe_redirect/2`

dweremeichik:
`render_upload` can emit errors that are not redirects. `upload/4` now throws nice errors for the following `render_upload` errors: [:too_large, :too_many_files, :not_accepted]. We let new/unknown errors simply fail at the case match so they can be implemented in the future. Sessions and redirects are passed through to `maybe_redirect` as before.

Fixes #167

### DIFF
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -273,7 +273,56 @@ defmodule PhoenixTest.Live do
     session.view
     |> file_input(form.selector, live_upload_name, [entry])
     |> render_upload(file_name)
+    |> maybe_throw_upload_errors(session, file_name, live_upload_name)
     |> maybe_redirect(session)
+  end
+
+  defp maybe_throw_upload_errors({:error, [[_id, error]]}, session, file_name, live_upload_name) do
+    case error do
+      :not_accepted -> raise ArgumentError, message: not_accepted_error_msg(session, file_name, live_upload_name)
+      :too_many_files -> raise ArgumentError, message: too_many_files_error_msg(session, file_name, live_upload_name)
+      :too_large -> raise ArgumentError, message: too_large_error_msg(session, file_name, live_upload_name)
+    end
+  end
+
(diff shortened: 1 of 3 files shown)


lib/phoenix_test/live.ex
```diff
@@ -273,9 +273,58 @@ defmodule PhoenixTest.Live do
...
+  end
+
+  defp too_large_error_msg(session, file_name, live_upload_name) do
+    %{name: name, max_file_size: max_file_size} = session.conn.assigns.uploads[live_upload_name]
+
+    """
+    File too large.
+
+    While attempting to upload "#{file_name}", you've exceeded the maximum file size of #{max_file_size} bytes. If this is intentional,
+    consider updating `allow_upload(:#{name}, max_file_size: #{max_file_size})`.
```
germsvel, line 324: Love the error messages! 🥳

germsvel (approved):
Looks good to me. Thanks for doing this! (sorry it took a while to review)

PR #179, opened 2025-02-07
Docs: Mention playwright driver

ftes:


germsvel (approved):
Thanks for doing this! 🥳 

I was trying to do this while adding docs to the Driver, but just realized this can go in without the Driver docs.

ftes:
Thanks, I appreciate it!

PR #180, opened 2025-02-10
Include email input fields in form data

szajbus:
Include email fields in form data


germsvel (approved):
Thanks for making this change! 🎉

Issue #132, opened 2024-10-01
fill_in without a label


andrewtimberlake:
Sometimes a form element doesn’t have a label:
![CleanShot 2024-10-01 at 12 01 15@2x](https://github.com/user-attachments/assets/4b468587-6dd4-4b2d-9a31-f8aa1acdd5c0)

It would be nice to be able to use `fill_in("model[field]", nil, with: "text")` or similar.

germsvel:
@andrewtimberlake in those cases, it's good to have a hidden label (at least as far as I understand accessibility, but I'm not an expert). So what I'd recommend is have a hidden label (with tailwind you could add `sr-only` class.)

This is from [mdn docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label): 

> The label text is not only visually associated with its corresponding text input; it is programmatically associated with it too. This means that, for example, a screen reader will read out the label when the user is focused on the form input, making it easier for an assistive technology user to understand what data should be entered.

That would allow you to target it with `fill_in("Domain", with: "text")`

andrewtimberlake:
Thanks, that’s what I ended up doing ;-)

slashmili:
Would you be open to adding reasoning in the documentation? I've been trying to find a solution, and while I've used `hidden`, I've always thought of it as a hack! Indeed making the label `sr-only` is much more appropriate solution

germsvel:
@slashmili I would love that in the docs. You up for opening a PR to include that?

PR #188, opened 2025-03-17
update docs for fill_in

slashmili:
to explain the reason behind using labels.

as discussed in https://github.com/germsvel/phoenix_test/issues/132


germsvel (approved):
Thanks @slashmili! 🥳

PR #195, opened 2025-03-29
`assert_path` handles query params with []

### LINKED ISSUE #168: `query_params` option for `assert_path/3` does not handle array values correctly
When calling assert_path as like this on working code:
```
assert_path(~p"/my-path/", query_params: %{"my_key" => ["1", "2"]}
```
the assertion is erroring with the following error:
```
Expected query params to be "my_key=12" but got "my_key[]=1&my_key[]=2"
```
Looking further into it, it looks like `PhoenixTest.Utils.stringify_keys_and_values/1` is the culprit.

I propose a fix where when the value of the query params is a list, it returns the params in the url array style.

After connecting with German over this, I plan to put in a PR to address this.

germsvel:
Resolves #168

What changed?
============

We improve `assert_path` to handle query params that have lists.

There are many cases when we want to assert the query includes things like `users[]=frodo&users[]=sam`.

The assertion would present those as a list in the map:

```elixir
assert_path("/some/path", query_params: %{users: ["frodo", "sam"]})
```

This commit makes it so that we can do that.

### DIFF
--- a/lib/phoenix_test/assertions.ex
+++ b/lib/phoenix_test/assertions.ex
@@ -238,12 +238,12 @@ defmodule PhoenixTest.Assertions do
     params = Utils.stringify_keys_and_values(params)
 
     uri = URI.parse(PhoenixTest.Driver.current_path(session))
-    query_params = uri.query && URI.decode_query(uri.query)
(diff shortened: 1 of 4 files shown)


idlehands:
Thank you @germsvel!

PR #196, opened 2025-03-31
Allow empty maps to be passed into assert_path as query_params

TheBoysAreZachInTown:
Initially, there was no way to assert that a path had no query params. If you passed into `assert_path` `query_params: ""` or `query_params: nil` you would get: `(FunctionClauseError) no function clause matching in PhoenixTest.Utils.stringify_keys_and_values/1` This is because `Utils.stringify_keys_and_values` is expecting a map.

If you passed in `query_params: %{}`, you would get: `Empty map results in "Expected query params to be "" but got nil.` This is because URI.parse returns `nil` when a query param is not present.

The fix is to assert true if either the query_params match the passed in expected params or if the parsed query_params are `nil` and the expected params are an empty map.


germsvel (approved):
I like the change. Thanks for adding this @TheBoysAreZachInTown!

PR #198, opened 2025-04-01
feat: introduce `click` helper for liveviews

StephanH90:
# Overview

Adds a `click/2` helper which can be used to click on elements that have a `phx-click` attribute.

Will raise an error when being used in a static view.

## Background
It is not realistic that we have a label in every button when we can use `aria-label` for this very purpose. This helper allows you to write more meaningful tests:

```html
<!-- You can write this -->
<button aria-label="toggle-menu" phx-click="toggle-menu">
  <.icon name="some-icon" />
</button>

<!-- Instead of this -->
<button phx-click="toggle-menu">
  <span class="sr-only">toggle menu</span>
  <.icon name="some-icon" />
</button>

<!-- you can also click div elements but that is somewhat debatable -->
<div phx-click="toggle-menu">
  <.icon name="some-icon" />
</div

```

### DIFF
--- a/lib/phoenix_test.ex
+++ b/lib/phoenix_test.ex
@@ -218,6 +218,30 @@ defmodule PhoenixTest do
   end
 
   @doc """
+  Clicks an element with the given selector. The clicked element needs to have
+  a `phx-click` attribute defined. However, the click helper has several limitations:
+
+  - it does not work for Static pages as it does not make sense in the context of a static view.
+  - it does not work with form elements (use the designated helpers like `check` etc. for that)
+
+  ## Examples
+
+  ```heex
+  <div phx-click="toggle" aria-label="toggle element">Element that toggles something</div>
+  <button class="toggle-panel" data-element-id="1">Element with class and data-element attribute</button>
+  ```
+
+  ```elixir
+  session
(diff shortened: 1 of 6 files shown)


germsvel:
@StephanH90 thanks for opening this PR, and even more for offering the **background** section! 🥳 

As you point out, we could use `aria-labels` for things that we currently do not. I would love to support `aria-labels`. 

So if you have something like what you showed: 

```html
<button aria-label="toggle-menu" phx-click="toggle-menu">
  <.icon name="some-icon" />
</button>
```

You can click that with `click_button("toggle-menu")`. 

☝️ that's the work that I'd love to happen in PhoenixTest. 

But I do not want to introduce a `click` that only works with LiveView's just so we can click on `phx-click`s on anything. So, I'm going to close this PR. 

If you're interested in doing some work for the `aria-label`, let's do that! We should open an issue and get a plan together.

StephanH90:
You can specify how you would like the API to look like. I can implement it. I don't feel like writing a great deal about 80 lines of code. I have no feelings about what it should look like at all. I wrote this code in the train on my way to work.

If you aren't interested, that's ok too, I will just use my own fork.

PR #201, opened 2025-04-04
Fill hidden input (mimic JS hook)

ftes:
Allow setting hidden input values.
My use case: Avoid switching to browser based test just because there is some small JS hook controlling a hidden input, while the user interacts with a visible input that has no `name` (is not submitted).

  ### Example

  Given the following:

  ```html
  <input id="monetary-hidden" type="hidden" name="monetary" value="200" />
  <label>Monetary amount <input value="2.00" phx-hook="Monetary" id="monetary-presentation" /></label>
  ```

  We can fill in the `monetary-hidden` field:

  ```elixir
  session
  |> fill_in("Monetary amount", with: "42.00", with_hidden: {"monetary-hidden", "4200"})
  ```

### Discussion
- Pro: Underlying data structure change is pretty generic, could be exposed in more ways in future.
- Pro: I think the with_hidden solution is pretty expressive for my use case. Still validates there is a visible, non-disabled input for the user.
- Con: It's not very generic, so unlikely to fit other (similar) use cases. E.g. only fill_in is supported atm, no other input functions.
- Alternative: add `fill_in_hidden` function. More flexible, less guarantees for my use case.
- Alternative: add `additional_data` arg to `submit`. More flexible, less guarantees for my use case.
- Alternative: add `ignore_label` option to `fill_in`. Caller must also pass hidden input ID selector as `input_selector` (which overrides the default `input:not([type=hidden])`. Not expressive at all.

### Under the hood
- Add `ActiveForm.force_form_data: %FormData{}`
  - live: passed to `render_submit` as extra `value` arg
  - static: merged with regular `form_data`

### DIFF
--- a/lib/phoenix_test.ex
+++ b/lib/phoenix_test.ex
@@ -463,7 +463,7 @@ defmodule PhoenixTest do
 
   - `exact`: whether to match label text exactly. (Defaults to `true`)
 
-  - `hidden_input_id`: force hidden input data.
+  - `with_hidden`: mimic JavaScript hook that sets hidden input. Format: `{hidden_input_id, value}`
 
   ## Examples
 
@@ -511,6 +511,28 @@ defmodule PhoenixTest do
   |> fill_in("Name", with: "Aragorn", exact: false)
   ```
 
+  ## Hidden inputs and JavaScript hooks
+
+  If your view relies on JavaScript hooks, consider using a browser based solution
+  such as `phoenix_test_playwright`. However, for small edge cases you can mimic
+  JS hooks using the `with_hidden` option.
+
+  ### Example
+
+  Given the following:
+
+  ```html
+  <input id="monetary-hidden" type="hidden" name="monetary" value="200" />
+  <label>Monetary amount <input value="2.00" phx-hook="Monetary" id="monetary-presentation" /></label>
+  ```
+
+  We can fill in the `monetary-hidden` field:
+
+  ```elixir
+  session
+  |> fill_in("Monetary amount", with: "42.00", with_hidden: {"monetary-hidden", "4200"})
+  ```
+
   ## Labels and Accessibility
 
   This function requires a label to target the input element. This is by design, as labels are
@@ -526,7 +548,7 @@ defmodule PhoenixTest do
   ```
   """
   def fill_in(session, label, opts) when is_binary(label) and is_list(opts) do
-    opts = Keyword.validate!(opts, [:with, :hidden_input_id, exact: true])
+    opts = Keyword.validate!(opts, [:with, :with_hidden, exact: true])
     Driver.fill_in(session, label, opts)
   end
 
@@ -565,7 +587,7 @@ defmodule PhoenixTest do
   ```
   """
   def fill_in(session, input_selector, label, opts) when is_binary(label) and is_list(opts) do
-    opts = Keyword.validate!(opts, [:with, :hidden_input_id, exact: true])
+    opts = Keyword.validate!(opts, [:with, :with_hidden, exact: true])
     Driver.fill_in(session, input_selector, label, opts)
   end
 
--- a/lib/phoenix_test.ex
+++ b/lib/phoenix_test.ex
@@ -463,6 +463,8 @@ defmodule PhoenixTest do
 
   - `exact`: whether to match label text exactly. (Defaults to `true`)
 
+  - `hidden_input_id`: force hidden input data.
+
   ## Examples
 
   Given we have a form that contains this:
@@ -524,7 +526,7 @@ defmodule PhoenixTest do
   ```
   """
   def fill_in(session, label, opts) when is_binary(label) and is_list(opts) do
-    opts = Keyword.validate!(opts, [:with, exact: true])
+    opts = Keyword.validate!(opts, [:with, :hidden_input_id, exact: true])
     Driver.fill_in(session, label, opts)
   end
 
@@ -563,7 +565,7 @@ defmodule PhoenixTest do
   ```
   """
   def fill_in(session, input_selector, label, opts) when is_binary(label) and is_list(opts) do
-    opts = Keyword.validate!(opts, [:with, exact: true])
+    opts = Keyword.validate!(opts, [:with, :hidden_input_id, exact: true])
     Driver.fill_in(session, input_selector, label, opts)
   end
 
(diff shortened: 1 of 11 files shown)


lib/phoenix_test/active_form.ex
```diff
@@ -3,7 +3,13 @@ defmodule PhoenixTest.ActiveForm do
 
   alias PhoenixTest.FormData
 
-  defstruct [:id, :selector, form_data: FormData.new(), uploads: FormData.new()]
+  defstruct [
+    :id,
+    :selector,
+    form_data: FormData.new(),
+    force_form_data: FormData.new(),
```
ftes, line 10: Alternative name: `additional_data`?

lib/phoenix_test.ex
```diff
@@ -463,6 +463,8 @@ defmodule PhoenixTest do
 
   - `exact`: whether to match label text exactly. (Defaults to `true`)
 
+  - `with_hidden`: mimic JavaScript hook that sets hidden input. Format: `{hidden_input_id, value}`
```
ftes, line 466: Unsure about the `with_hidden: {id, value}` format. Its concise, but not very readable.

germsvel:
@ftes just looking at this now. I understand the desire to not need to use JS for a test like this. But I'm always hesitant to break the encapsulation PhoenixTest's API has for one-off use cases. I think then everyone will want a slight tweak to the API to satisfy their own use case. 

That was the case with supporting LiveView's async operations. People wanted me to introduce a quick map to `render_async`, but it would have broken PhoenixTest's API and introduced a function that only works for LiveViews. I think the `:timeout` approach that we ended up landing on (though a _ton_ more work) was a better solution. 

But I like where your heads at regarding: can we do some extra stuff with PhoenixTest to delay the need to introduce JS? I've been thinking a little about that with JS hooks like `show/hide`. It would be cool if we could test that with PhoenixTest (example: show/hide modals with JS). But I'd like to think of a more holistic approach. 

Right now, I'm not sure what that approach would be. And I think at least we have the `#unwrap` escape hatch. But I'm open to hearing other thoughts.

ftes:
> I've been thinking a little about that with JS hooks like show/hide

That would be neat. Given the `%JS{}` abstraction that should be feasible.

> at least we have the #unwrap escape hatch

My problem with that is that it's either or. Either `PhoenixTest` form handling OR custom. But not both interleaved.
Not possible: `|> fill_in("Label", with: "value") |> unwrap(&set_hidden_input/1) |> submit()`.
Unless I break encapsulation I
- cannot add to form data for future `change/submit` callbacks
- cannot get existing form data and include to include in a `render_change` or `render_submit` call

Does that make sense?

germsvel:
> My problem is that it's either or. Either PhoenixTest form handling start to finish (fill + submit) OR custom form handling. But not interleaved.

Yeah, that makes total sense. I wonder if we could extend `#unwrap` to pass several arguments depending on the anonymous function provided. 

So, something like this (pseudocode): 

```elixir
## For LiveViews

# only one argument takes the view
|> unwrap(fn view -> ... end)

# two arguments take view and conn
|> unwrap(fn view, conn -> ... end)

# three arguments take in view, conn, and ActiveForm turned into payload
|> unwrap(fn view, conn, data -> ... end)
```

That would allow you to do all the fill_in stuff if needed, and then `#unwrap`, add additional data, and submit it: 

```elixir
session
|> fill_in("Name", with: "Frodo")
|> unwrap(fn view, _conn, data -> 
   view
   |> form("form", data)
   |> render_submit(additional_data_you_want_to_pass)
end)
```

My only concern with that is that the `data` becomes something of a public API, so if we change something about it, we could cause breaking changes. But maybe that's not as big a risk as I'm thinking.

ftes:
> data becomes something of a public API

Don't Phoenix/Phoenix.LiveView basically provide the API spec by defining in which shape they expect data for `dispatch/render_change/render_submit`? In that case it seems fine. Or are there maybe some ambiguities in how to represent data, e.g. for nested lists (has/embeds_many)?

> `|> unwrap(fn view, conn, data -> ... end)`

That would definitely be an improvement!

Still less ideal for my use case than `fill_in_hidden`. Because
1. it doesn't validate there is indeed a hidden field and
2. the hidden value isn't preserved in further calls - have to unwrap everything from that point on if I need the hidden value to persist

Unlike the `unwrap` extension you proposed, `fill_in_hidden` seems to me less opinionated because it operates at the DOM/HTML abstraction level.  Hidden fields and how they are handled/submitted in forms are part of the HTML spec. Whereas `unwrap` and how the data is passed is rather technically necessitated by the `PhoenixTest` architecture.

germsvel:
> Unlike the unwrap extension you proposed, fill_in_hidden seems to me less opinionated because it operates at the DOM/HTML abstraction level.

Yeah, I think it would be nice to find a way to handle those hidden fields. I was just saying that leaning on `#unwrap` until we figure out the best abstraction gives us flexibility to _not_ introduce APIs that require breaking changes later. 

But, I'm game to try to find that abstraction right now. 

---

I like your example where the value you want to submit to the backend is different from the presented value (e.g. money). 

Could we also use `fill_in_hidden` for things like the old-but-classic [select2](https://select2.org/getting-started/basic-usage)? Or would that require something else like a `select_hidden`?

ftes:
> Could we also use fill_in_hidden for things like the old-but-classic [select2](https://select2.org/getting-started/basic-usage)? Or would that require something else like a select_hidden?

I've never used `select2` before. I briefly read the docs now.
For simple `<select>` tags (static `<option>`s) you could probably use `PhoenixTest.select` as is, because `select2` just "progressively enhances" the `<select>` input and synchronizes the selected state with the original `<select>` node.

`select2` doesn't seem to use hidden inputs at all, so `fill_in_hidden` is likely not useful here.

The same is true for `tom-select`, which in the more basic use cases replaces an existing `select` node.

I'd have to try out the more advanced use cases to say for sure (multiple selection, dynamic data fetching, dynamic creation of options by user).

ftes:
FYI I started preparing some examples to explore this idea further:
> If you render regular form inputs in your view and design your JS hooks to progressively enhance them, you can use regular PhoenixTest input functions in your tests.

https://github.com/ftes/phoenix_test_playground/blob/main/test/my_app_web/features/blog_test.exs

Seems to work well for `tom-select` (combobox) and `quill` (rich text).
Though for quill I probably would want to store either markdown or HTML alongside the document model (delta).
Either I'd want a second, hidden, input for that. Or bundle delta + HTML in a JSON serialized string so it maps 1:1 to a vanilla input.

ftes:
Closing this, it's been too long :)

PR #210, opened 2025-04-25
Fix: Find select nested in label

ftes:
### Given
```html
      <label>
        Name
        <select id="name" name="name">
          <option value="select_1">Select 1</option>
          <option value="select_2">Select 2</option>
        </select>
      </label>
```

### when
```ex
|> select("Name", option: "Select 1")
```

### then
option is selected.

## Why does it currently fail?
The option text is included in `Floki.text` when searching for the label by text.
So an exact match fails, because `"Name" != "Name Select 1 Select 2"`.


germsvel (approved):
👍 thanks for catching this!

PR #224, opened 2025-06-16
Fix: Repeated select/3 calls for multi select

ftes:
Calling `select/3` multiple times shouldn't fail.

For multi selects, it currently fails.

I've only created a failing test case for now.


germsvel:
Thanks for getting some tests going @ftes!

germsvel:
Closing in favor of #251

PR #225, opened 2025-06-16
Fix: check/3 in group loses initial checked

ftes:
Regression in phoenix_test 0.7

Only created a failing test case for now.


germsvel:
Thanks for getting some tests going @ftes

PR #231, opened 2025-07-04
Dynamic endpoint

jpyamamoto:
Hello,

At my company we recently faced the challenge that we wanted to use this library for the tests of two different apps in the umbrella, each with a different endpoint.

This turned out to be unachievable with the current state of the library due to the endpoint configuration needing to be given at compile time.

In order to make it work, we had to hack into the library a way to retrieve the endpoint at runtime. We propose the following syntax:

```elixir
config :phoenix_test,
  endpoints: %{
    opt_app1: MyApp1.Endpoint,
    otp_app2: MyApp2.Endpoint
    ...
  }
```

In order to make it work, the `live` and `follow_redirect` macros defined by Phoenix had to be replicated in the library as functions that invoke `dispatch/5` instead of `get/3`, to manually pass the proper endpoint.

This MR preserves backwards compatibility as it will use the already established `:phoenix_test, :endpoint` config when `endpoints` isn't there.

Please, let me know if this would be a valuable contribution to the library and if so, if there are any changes I should make before this is merged.

### DIFF
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -17,28 +17,15 @@ defmodule PhoenixTest.Live do
   alias PhoenixTest.LiveViewTimeout
   alias PhoenixTest.Locators
   alias PhoenixTest.Query
-  alias PhoenixTest.Utils
+  alias PhoenixTest.EndpointHelpers
 
   defstruct view: nil, watcher: nil, conn: nil, active_form: ActiveForm.new(), within: :none, current_path: ""
 
   def build(conn) do
-    {:ok, view, _html} = live_with_current_endpoint(conn)
+    {:ok, view, _html} = EndpointHelpers.live_with_current_endpoint(conn)
     current_path = ConnHandler.build_current_path(conn)
     {:ok, watcher} = start_watcher(view)
     %__MODULE__{view: view, watcher: watcher, conn: conn, current_path: current_path}
-  end
-
-  defp live_with_current_endpoint(conn, path \\ nil, opts \\ []) do
-    cond do
-        is_binary(path) ->
-          Phoenix.LiveViewTest.__live__(dispatch(conn, Utils.current_endpoint(), :get, path), path, opts)
-
-        is_nil(path) ->
-          Phoenix.LiveViewTest.__live__(conn, nil, opts)
-
-        true ->
-          raise RuntimeError, "path must be nil or a binary, got: #{inspect(path)}"
-      end
   end
 
   defp start_watcher(view) do
@@ -271,11 +258,9 @@ defmodule PhoenixTest.Live do
       type: mime_type
     }
 
-    builder = fn -> Phoenix.ChannelTest.__connect__(Utils.current_endpoint(), Phoenix.LiveView.Socket, %{}, []) end
-
     upload_progress_result =
       session.view
-      |> Phoenix.LiveViewTest.__file_input__(form.selector, live_upload_name, [entry], builder)
+      |> EndpointHelpers.file_input_with_current_endpoint(form.selector, live_upload_name, [entry])
       |> render_upload(file_name)
       |> maybe_throw_upload_errors(session, file_name, live_upload_name)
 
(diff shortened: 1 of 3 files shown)


lib/phoenix_test/live.ex
```diff
@@ -259,9 +271,11 @@ defmodule PhoenixTest.Live do
       type: mime_type
     }
 
+    builder = fn -> Phoenix.ChannelTest.__connect__(Utils.current_endpoint(), Phoenix.LiveView.Socket, %{}, []) end
+
     upload_progress_result =
       session.view
-      |> file_input(form.selector, live_upload_name, [entry])
+      |> Phoenix.LiveViewTest.__file_input__(form.selector, live_upload_name, [entry], builder)
```
germsvel, line 278: Did we also need this change as part of the endpoint PR? Or is this done for a different reason?

lib/phoenix_test/live.ex
```diff
@@ -259,9 +271,11 @@ defmodule PhoenixTest.Live do
       type: mime_type
     }
 
+    builder = fn -> Phoenix.ChannelTest.__connect__(Utils.current_endpoint(), Phoenix.LiveView.Socket, %{}, []) end
```
germsvel, line 274: So I get a better sense of whether or not it makes sense to include this, could you point me to where you got this code from? Was it from `file_inputs`'s implementation?

lib/phoenix_test/utils.ex
```diff
@@ -13,4 +15,28 @@ defmodule PhoenixTest.Utils do
         {to_string(k), to_string(v)}
     end)
   end
+
+  def current_endpoint() do
+    current_app = Mix.Project.config()[:app]
+
+    if is_nil(current_app) do
+      raise "no :app set in Mix project config"
```
germsvel, line 23: What do you think about having a nicer error message here? If a user gets this error without knowing that it's due to the config, it might not be clear _why_ they need to set the `:app` and for which library. Maybe we could suggest something here, or let them know what we expected? 

```suggestion
      raise "could not find `:my_app, :phoenix_test` set in project config. Please set it along with your endpoint."
```

I'm not sure if that's too verbose, but maybe something like that?

lib/phoenix_test/utils.ex
```diff
@@ -13,4 +15,28 @@ defmodule PhoenixTest.Utils do
...
+    end
+
+    Application.fetch_env(:phoenix_test, :endpoints)
+    |> case do
+      {:ok, endpoints} -> 
+        Access.fetch(endpoints, current_app)
+        |> case do
+          {:ok, endpoint} -> endpoint
+          :error -> raise "no endpoint set for #{current_app} in config"
```
germsvel, line 32: Same thought as above. Could we add more clarity as to which library is raising this error and what's expected?

lib/phoenix_test/utils.ex
```diff
@@ -13,4 +15,28 @@ defmodule PhoenixTest.Utils do
...
+        end
+
+      :error ->
+        case @endpoint do
+          nil -> raise "no endpoint set in config"
```
germsvel, line 37: Same question as above. Could we improve this error message?

lib/phoenix_test/live.ex
```diff
@@ -17,18 +17,30 @@ defmodule PhoenixTest.Live do
...
+  defp live_with_current_endpoint(conn, path \\ nil, opts \\ []) do
+    cond do
+        is_binary(path) ->
+          Phoenix.LiveViewTest.__live__(dispatch(conn, Utils.current_endpoint(), :get, path), path, opts)
+
+        is_nil(path) ->
+          Phoenix.LiveViewTest.__live__(conn, nil, opts)
+
+        true ->
+          raise RuntimeError, "path must be nil or a binary, got: #{inspect(path)}"
```
germsvel, line 40: This might be a nit, but maybe `ArgumentError` would be more appropriate here?

lib/phoenix_test/live.ex
```diff
@@ -17,18 +17,30 @@ defmodule PhoenixTest.Live do
...
 
   def build(conn) do
-    {:ok, view, _html} = live(conn)
+    {:ok, view, _html} = live_with_current_endpoint(conn)
     current_path = ConnHandler.build_current_path(conn)
     {:ok, watcher} = start_watcher(view)
     %__MODULE__{view: view, watcher: watcher, conn: conn, current_path: current_path}
   end
 
+  defp live_with_current_endpoint(conn, path \\ nil, opts \\ []) do
```
germsvel, line 31: What do you think about moving all of these new "shim" functions into their own module? I think they could live there along with what is currently in `Utils.current_endpoint()`. It might be nice to have a single place where we're doing all this patching so that we can handle that. 

Naming is hard, of course, so I can't think of something really good, but maybe something like `PhoenixTest.EndpointHelpers`? I'm 100% open to ideas here. What do you think?

germsvel (commented):
Thanks for opening this PR! This has been something that people have requested in the past, but I never had the time to try to figure out how we could make it happen. 

I've given this a look, and I think the general approach is okay. I just have a few comments/suggestions that I think would clean things up a bit (and a question out of ignorance 😄 )

If anything is unclear, let me know! Happy to try and guide things a bit more.

lib/phoenix_test/live.ex
jpyamamoto: Yes, you can find the implementation here: https://github.com/phoenixframework/phoenix_live_view/blob/4aa2242f72dcba5b800a72da36cb5c170879d538/lib/phoenix_live_view/test/live_view_test.ex#L1216

jpyamamoto:
Modified the error messages, let me know if they're better this way.

siraj-samsudeen:
ok German - i would wait for your other PR to take shape to see whether this fix is even needed. I have created an issue in the mix_test_interactive repo also, to see whether there is any other way to handle this - [mix test.interactive fails on a test that mix test passes using PhoenixTest · Issue #145 · randycoulman/mix_test_interactive](https://github.com/randycoulman/mix_test_interactive/issues/145)

PR #235, opened 2025-07-18
Fix phx-trigger-action handling for dynamically rendered forms

dkulchenko:
**Problem:**
PhoenixTest failed to handle forms with `phx-trigger-action` that are rendered dynamically after LiveView events (not present in initial dead render). This resulted in "Could not find element with selector" errors during form submission.

**Root Cause:**
The `maybe_redirect/2` function in `PhoenixTest.Live` was using `session.conn.resp_body` (stale HTML from initial page load) when switching to Static context to handle `phx-trigger-action` forms. Dynamically rendered forms only exist in the current LiveView HTML, not in the original response body.

**Solution:**
Update `maybe_redirect/2` to use the current LiveView HTML as the `resp_body` when creating the Static session. This ensures `PhoenixTest.Static.submit_form/3` can find dynamically rendered forms.

Changes:
- `session.conn` → `%{session.conn | resp_body: html}`

**Test Case Added:**
- New `DynamicFormLive` that renders a form only after button click
- Test case: "handles phx-trigger-action on dynamically rendered forms"
- Validates the fix and prevents regressions

Fixes the core issue where PhoenixTest looked for forms in wrong HTML context.

### DIFF
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -475,7 +475,7 @@ defmodule PhoenixTest.Live do
         active_form? = form.selector == active_form.selector
         form_data = FormData.merge(form.form_data, if(active_form?, do: active_form.form_data, else: FormData.new()))
 
-        session.conn
+        %{session.conn | resp_body: html}
         |> PhoenixTest.Static.build()
         |> PhoenixTest.Static.submit_form(form.selector, form_data)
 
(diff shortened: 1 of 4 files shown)


germsvel (approved):
@dkulchenko love the PR description and the implementation! 

Thanks for fixing this! 🥳

Issue #237, opened 2025-07-18
`uncheck/3|4` on checkboxes with `phx-click` does not pass along values in `phx-value`


joeytrapp:
`check/3|4` and `uncheck/3|4` differ in their behavior in that a checkbox with a `phx-click` attribute will correctly pass along the values specified in any `phx-value` attributes (i.e. `phx-value-id={@record.id}`), but the corresponding `uncheck/3|4` call sets the params map to `%{}`. The error we see as a result is a function clause error since we’re pattern matching on the incoming params.

Here is a phoenix_playground demonstration that provokes the issue:
https://gist.github.com/joeytrapp/d31b4df0559fd3d9f43d58b8841a6680

PR #246, opened 2025-08-27
support phx-value and JS command values with uncheck

joeytrapp:
Closes: https://github.com/germsvel/phoenix_test/issues/237

Sorry of the delay, had to work through submission approval.

This PR updates `uncheck/3|4` to check if the checkbox has a `phx-click` attribute and either also has `phx-value` attributes or the `phx-click` attribute is a JS command with a value assigned, and makes sure those values are passed as params to the `handle_action/3` callback.

If there is a `phx-click` attribute but no identifiable values, the previous behavior of using an empty map for params is maintained for compatibility.


germsvel (approved):
Thanks @joeytrapp! Love the change.

PR #248, opened 2025-08-28
Preserve order of appearance for array inputs

soundmonster:
Elements in lists/arrays in forms (referenced by unindexed `field_name[]`) have been aggregated in the order they appear in the HTML document before the refactoring in https://github.com/germsvel/phoenix_test/commit/b1400b3faee35f9e5f078422042b309895c00e51
Since then, the order has been the opposite.
This works in most cases, but sometimes the order is significant and flipping it can become problematic.

This PR reverts the order to the "natural" order of occurrence in the document.

### DIFF
--- a/lib/phoenix_test/form_data.ex
+++ b/lib/phoenix_test/form_data.ex
@@ -40,7 +40,7 @@ defmodule PhoenixTest.FormData do
           if value in existing_value do
             existing_value
           else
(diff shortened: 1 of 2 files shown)


germsvel (approved):
Thanks @soundmonster! Happy to get this in.

PR #249, opened 2025-09-16
Add `change_form/3` and `submit_form/3` as wrappers for LiveViewTest's `form/3`

ericridgeway:
Much of the generated tests you get with Phoenix LiveView have blocks like
  ```elixir
  valid_attrs = %{name: "hello", favorite_color: "blue"}

  assert view
    |> form("#my_form", user: valid_attrs)
    |> render_submit() =~ "Name updated"
  ```

Notice they aren't filling in each field manually, but instead using a map of attrs meant to mimic what the user will actually be submitting when they send the form

This is especially useful when filling in many form fields at once

These two functions let you use that familiar Phoenix.LiveViewTest style
  ```elixir
  conn
  |> visit("/")
  |> change_form("#my_form", user: @valid_attrs)
  ```

### DIFF
--- a/lib/phoenix_test.ex
+++ b/lib/phoenix_test.ex
@@ -573,6 +573,61 @@ defmodule PhoenixTest do
   end
 
   @doc """
+  Change a form using map syntax
+
+  An alternative to `fill_in/3`, which wraps `Phoenix.LiveViewTest.form/3` followed
+  by `Phoenix.LiveViewTest.render_change/1`
+
+  Often you'll have multiple field_name & field_value's you want to fill a form in with, but
+  don't want to manually use `fill_in/3` for each of them
+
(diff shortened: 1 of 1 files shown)


germsvel:
@ericridgeway thanks for opening this PR. PhoenixTest used to have an API that supported something like that. But I made a design decision to move away from that so we could better support finding form helpers by labels. 

There's a little bit more of an explanation in the upgrade docs for v0.2.13: 

https://hexdocs.pm/phoenix_test/upgrade_guides.html#upgrading-to-0-2-13

ericridgeway:
@germsvel That... makes a ton of sense. Really good point about decoupleing the html tests from the shape of a behind-the-scenes changeset

Thanks for the reply and the INCREDIBLE library. It makes writing frontend tests quick and doable instead of a chore that often gets ignored. I wish it was core Phoenix <3

PR #254, opened 2025-09-20
Fix #227

### LINKED ISSUE #227: click_button finds multiple elements outside of `within`
Hi there 👋 

I'm getting the error:
```
(ArgumentError) selector "button[phx-click=\"play\"]" returned 2 elements and 2 of them matched the text filter "Play"
```

My code (simplified selectors):

```elixir
      within(session, "li:has(div:fl-contains('some text'))", fn session ->
        session |> click_button("button[phx-click=\"play\"]", "Play")
      end)
```

Of the two matched elements it shows me in the error message, only one of them is within the `within`. (Confirmed via inspecting `PhoenixTest.Live.render_html(session)` inside the within).

So it seems like the click_button is run on the full html.

----

I've browsed around in the source and found the potential source of the error:

```elixir
  def click_button(session, selector, text) do
    html = render_html(session)
    button = Button.find!(html, selector, text)

    cond do
      Button.phx_click?(button) ->
        session.view
        |> element(selector, text)
        |> render_click()
        |> maybe_redirect(session)
```

Should it maybe be: `element(button.selector, text)`?
Since `selector` is scoped to the sessions within, but `session.view` probably isn't?

---
aside: the fl-contains filter was hard to find, it would be easier if the within would accept another argument as a text filter

spicychickensauce:
I took your test as a starting point and implemented a possible fix for https://github.com/germsvel/phoenix_test/issues/227

I made the minimal changes needed to fix this issue, however I think a proper fix would affect a lot more code.
Because of that I kept this PR in a draft state, as I don't want to do many invasive changes.

Essentially everywhere a struct like `Button` or e.g. `Field` is used, the constructed selector needs to respect the current `within`. So this bug is almost certainly also present outside of `click_button`.

Instead of what I did in my fix, a better approach would probably be to pass around the current session into `Button.find!` and similar such that the correct selector can be constructed then.

### DIFF
--- a/lib/phoenix_test/live.ex
+++ b/lib/phoenix_test/live.ex
@@ -82,6 +82,8 @@ defmodule PhoenixTest.Live do
   end
 
   defp handle_click_button(session, button) do
+    button = %{button | selector: PhoenixTest.SessionHelpers.scope_selector(session.within, button.selector)}
+
     cond do
       Button.phx_click?(button) ->
         session.view
--- a/lib/phoenix_test/session_helpers.ex
+++ b/lib/phoenix_test/session_helpers.ex
@@ -2,11 +2,14 @@ defmodule PhoenixTest.SessionHelpers do
   @moduledoc false
   def within(session, selector, fun) when is_binary(selector) and is_function(fun, 1) do
     session
-    |> Map.update!(:within, fn
-      :none -> selector
-      parent when is_binary(parent) -> parent <> " " <> selector
-    end)
+    |> Map.update!(:within, &scope_selector(&1, selector))
     |> fun.()
     |> Map.put(:within, :none)
+  end
+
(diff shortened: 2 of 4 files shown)


germsvel:
@spicychickensauce just took a look at this. Thanks for opening it! 

It looks mostly right to me. But, any chance you could move the `scope_selector` calls right above the code blocks where we call `render_click`? I'd like to keep the scoping to a minimum (in terms of surface area in the code).

spicychickensauce:
@germsvel Alright, I changed it as you suggested :+1: 
There were some places where I removed the `scope_selector` call completely (`upload` and `fill_in`), as the `field.selector` wasn't actually used.
Those probably do have this bug, but maybe better to address it once someone actually runs into it, as they work a bit differently

germsvel:
Thanks @spicychickensauce!

PR #280, opened 2025-11-20
feat: add support for phx-change event on input

StephanH90:
Currently, `phx-change` only works on form elements. This PR adds the functionality to also test this on inputs.


germsvel (approved):
Thanks for this work @StephanH90! 🎉 

A quick follow up question. This is something we would also want to support for checkboxes and radio buttons, right? 

It doesn't have to be done as part of this PR (we can open an issue for the others), but I just want to make sure so we can have consistent behavior across inputs.

PR #307, opened 2026-04-02
Add process-local endpoint configuration via `PhoenixTest.Config`

nbernardes:
### Problem

When using `phoenix_test`, the endpoint needs to vary depending on which test suite is running. In umbrella apps or projects with multiple Phoenix endpoints, `Application.get_env(:phoenix_test, :endpoint)` is global, so it can't differ per test suite and isn't safe when running tests in parallel. `put_endpoint/2` solves this for standard PhoenixTest (it sets conn.private), but external drivers like `phoenix_test_playwright` read the endpoint from Application config directly since they don't use `Plug.Conn`.

### Solution

Introduce `PhoenixTest.Config.put_endpoint/1` which stores the endpoint in the process dictionary. Since each `ExUnit` test runs in its own process, different test suites can set different endpoints without interfering with each other.

Endpoint resolution chain:
1. `conn.private[:phoenix_endpoint]` (per-conn)
3. Process dictionary (per-test-process) — new
5. `Application.get_env(:phoenix_test, :endpoint)` (global fallback)

`put_endpoint/2` now also sets the process-level config, so both approaches stay in sync.

 ### Usage

 ```elixir
   # In each ConnCase/FeatureCase:
   setup do
     PhoenixTest.Config.put_endpoint(MyAppWeb.Endpoint)
     {:ok, conn: Phoenix.ConnTest.build_conn()}
   end
 ```

 Fully backward-compatible - existing config :phoenix_test, :endpoint setups continue to work unchanged.

### DIFF
--- a/lib/phoenix_test.ex
+++ b/lib/phoenix_test.ex
@@ -49,10 +49,24 @@ defmodule PhoenixTest do
 
   PhoenixTest needs to know which endpoint to use for routing requests.
 
-  There are two ways of doing this:
+  There are three ways of doing this:
 
-  1. You can specify the endpoint at runtime in your `ConnCase` (or equivalent)
-  setup block:
+  1. You can set it per-test-process in your `ConnCase` (or equivalent) setup
+  block. This is the recommended approach for umbrella apps, multi-endpoint
+  setups, and compatibility with external drivers like `phoenix_test_playwright`:
+
+  ```elixir
+  setup do
+    PhoenixTest.Config.put_endpoint(MyAppWeb.Endpoint)
+    {:ok, conn: Phoenix.ConnTest.build_conn()}
+  end
+  ```
+
+  This stores the endpoint in the process dictionary, so it's parallel-safe -
+  different test suites can use different endpoints without interfering with
+  each other.
+
+  2. You can specify the endpoint on the conn directly in your setup block:
 
   ```elixir
   setup tags do
@@ -62,13 +76,17 @@ defmodule PhoenixTest do
   end
   ```
 
-  NOTE that this opens the option for umbrella apps to have different endpoints.
+  NOTE: `put_endpoint/2` also sets the process-level config, so external
+  drivers can resolve the endpoint too.
 
-  2. You can set it at compile time in `config/test.exs`:
+  3. You can set a global fallback in `config/test.exs`:
 
   ```elixir
   config :phoenix_test, :endpoint, MyAppWeb.Endpoint
   ```
+
+  NOTE: This is a global setting shared across all tests. For umbrella apps or
+  multi-endpoint setups, prefer option 1 or 2 above.
 
   ### Getting `PhoenixTest` helpers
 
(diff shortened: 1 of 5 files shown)


frm:
Hey @germsvel, is it possible to take a look at this? particularly useful for umbrellas with multiple endpoints

test/phoenix_test/config_test.exs
```diff
@@ -0,0 +1,54 @@
...
+  describe "endpoint!/0" do
+    test "returns endpoint when set via put_endpoint" do
+      Config.put_endpoint(PhoenixTest.WebApp.Endpoint)
+      assert Config.endpoint!() == PhoenixTest.WebApp.Endpoint
+    end
+
+    test "raises when no endpoint configured" do
+      Process.delete(:phoenix_test_endpoint)
+      original = Application.get_env(:phoenix_test, :endpoint)
+      Application.delete_env(:phoenix_test, :endpoint)
```
germsvel, line 43: I think this test makes it so we can't run this in `async: true` mode (since we're updating application config). Sadly, we may have to set `async: false`

lib/phoenix_test.ex
```diff
@@ -62,14 +76,18 @@ defmodule PhoenixTest do
   end
   ```
 
-  NOTE that this opens the option for umbrella apps to have different endpoints.
+  NOTE: `put_endpoint/2` also sets the process-level config, so external
+  drivers can resolve the endpoint too.
 
-  2. You can set it at compile time in `config/test.exs`:
+  3. You can set a global fallback in `config/test.exs`:
```
germsvel, line 82: I realize this was already set as `#2` (and we're just changing it to `#3`), but now that we're adding another option, it makes this seem less preferred. 

I still think the global config is the most common and simplest option. Since most apps aren't umbrella apps, I'd like to keep it as the "default" for people. So, I'd like to keep it as the first option. And only the people who _can't_ use it like that can look at what other options to opt into. 

So, could we set this as option 1? Then using the `put_endpoint` on the conn as second, and only mentioned the per-process `Config` one as third? 

That makes sense in my head because it goes from basic to complex: 

1. Single Phoenix app - use global
2. Umbrella app _without_ JS - use `put_endpoint` for conn
3. Umbrella app with JS and needs per-process stuff, use the new thing.

lib/phoenix_test/config.ex
```diff
@@ -0,0 +1,36 @@
+defmodule PhoenixTest.Config do
+  @moduledoc false
+
+  @process_key :phoenix_test_endpoint
+
+  def put_endpoint(endpoint) when is_atom(endpoint) do
```
germsvel, line 6: I think it'll be confusing that we have a `PhoenixTest.put_endpoint` and a `PhoenixTest.Config.put_endpoint`. Rather than do that, I wonder if we create a new function and call it `PhoenixTest.put_process_endpoint`. Perhaps it's a little to explicit, but it's (hopefully) clear. 

What do you think?

lib/phoenix_test/config.ex
```diff
@@ -0,0 +1,36 @@
...
+          end
+
+      Or set it on the conn directly:
+
+          {:ok, conn: Phoenix.ConnTest.build_conn() |> PhoenixTest.put_endpoint(MyAppWeb.Endpoint)}
+
+      Or set it globally in config/test.exs:
+
+          config :phoenix_test, :endpoint, MyAppWeb.Endpoint
+      """
```
germsvel, line 34: Can we also invert the order in which we show these messages? (see my other comment). 

I'd prefer it to be like this since it goes from simple & common case to more complex and less common (at least in my mind): 

1. global
2. build_conn |> put_endpoint
3. process put_endpoint

germsvel (commented):
Thanks so much for opening this PR! 🙏 

Sorry it took so long to review. I didn't even realize this was open (somehow). But I've also been very busy. So, thanks for the patience. 

I'm not opposed to introducing this. I don't typically work in Umbrella apps, but I can see what you mean by needing it with phoenix test playwright and umbrella apps. 

I left a few suggestions that I think would make it better. If something is unclear, let me know. Happy to chat more about it.

Issue #262, opened 2025-10-06
Support assert_has value finder for select


ftes:
Allow using

```ex
|> assert_has("select", label: "Race", value: "Elf")
```

for select inputs, similar to what already works for text inputs.

ftes:
> is it meant to assert that the select with label "Race" has an option with value "Elf"? Or is it saying that the selected option is the value "Elf"?

I would say the latter: `the selected option is the value "Elf"`.
Since this is a `select + value` assertion.
That is what I expect intuitively by looking at the function + option name. And by comparison with the behaviour of the `value` option when using with regular `input`s.
Maybe mention this explicitly in the docs?

For this case `the select with label "Race" has an option with value "Elf"`, I would expect to use an `option + value` assertion (not supported, don't think we need that).

ftes:
Closed by #309

PR #309, opened 2026-04-21
Support assert_has value finder for select

ftes:
Support `assert_has/3` and `refute_has/3` with `value:` for `select` elements by matching selected option text.

Implements #262.

### DIFF
--- a/lib/phoenix_test/assertions.ex
+++ b/lib/phoenix_test/assertions.ex
@@ -17,6 +17,7 @@ defmodule PhoenixTest.Assertions do
       :count,
       :exact,
       :label,
+      :selected,
       :text,
       :value
     ]
@@ -26,6 +27,7 @@ defmodule PhoenixTest.Assertions do
       count = Keyword.get(opts, :count, :any)
       exact = Keyword.get(opts, :exact, false)
       label = Keyword.get(opts, :label, :no_label)
+      selected = Keyword.get(opts, :selected, :no_selected)
       text = Keyword.get(opts, :text, :no_text)
       value = Keyword.get(opts, :value, :no_value)
 
@@ -34,6 +36,7 @@ defmodule PhoenixTest.Assertions do
         count: count,
         exact: exact,
         label: label,
+        selected: selected,
         text: text,
         value: value
       }
@@ -45,6 +48,7 @@ defmodule PhoenixTest.Assertions do
         count: opts.count,
         exact: opts.exact,
         label: opts.label,
+        selected: opts.selected,
         text: opts.text,
         value: opts.value
       ]
@@ -389,6 +393,7 @@ defmodule PhoenixTest.Assertions do
     "Expected #{count_elements(opts.count)} with #{inspect(selector)}"
     |> maybe_append_text(opts.text)
     |> maybe_append_value(opts.value)
+    |> maybe_append_selected(opts.selected)
     |> maybe_append_label(opts.label)
     |> append_found(found)
   end
@@ -397,6 +402,7 @@ defmodule PhoenixTest.Assertions do
     "Could not find #{count_elements(opts.count)} with selector #{inspect(selector)}"
     |> maybe_append_text(opts.text)
     |> maybe_append_value(opts.value)
+    |> maybe_append_selected(opts.selected)
     |> maybe_append_label(opts.label)
     |> maybe_append_position(opts.at)
     |> append_found_other_matches(selector, other_matches)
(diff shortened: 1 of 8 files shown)


lib/phoenix_test/element/form.ex
```diff
@@ -123,26 +123,9 @@ defmodule PhoenixTest.Element.Form do
...
-          end
-
-        {false, _} ->
-          [to_form_field(select, selected_options)]
-
-        _ ->
-          Enum.map(selected_options, &to_form_field(select, &1))
-      end
+      select
+      |> Html.selected_options()
```
ftes, line 127: Extracted

lib/phoenix_test/assertions.ex
```diff
@@ -462,14 +462,12 @@ defmodule PhoenixTest.Assertions do
   end
 
   defp value_finder_fun(value, selector, %Opts{} = opts) do
-    selector = selector <> "[value=#{inspect(value)}]"
-
     case opts.label do
       :no_label ->
-        &Query.find(&1, selector, Opts.to_list(opts))
+        &Query.find_by_value(&1, selector, value, Opts.to_list(opts))
```
ftes: Find by `value` implemented differently for `select` than `input`.

This makes we wonder: Should we make this explicit in the API: `assert_has("select", label: "Race", selected_option_text: "Elf")`?

lib/phoenix_test/html.ex
```diff
@@ -68,6 +68,18 @@ defmodule PhoenixTest.Html do
...
 
+  def selected_options(%LazyHTML{} = select) do
+    selected_options = all(select, "option[selected]")
+
+    if attribute(select, "multiple") do
+      selected_options
+    else
+      selected_options
+      |> Stream.concat(all(select, "option"))
+      |> Enum.take(1)
```
ftes: Too cryptic?
"Take first selected option, otherwise fall back to first option (selected by default)"

lib/phoenix_test/assertions.ex
germsvel: I think using one of the following would be good for clarity: 

```elixir
assert_has("select", label: "Race", selected: "Elf")

assert_has("select", label: "Race", selected_value: "Elf")

assert_has("select", label: "Race", selected_option: "Elf")
```

What do you think?

lib/phoenix_test/html.ex
germsvel: That is a bit cryptic.  I didn't realize that's what it was doing. 😄 

Can we add tests at this level that document that? I think I'd be okay with cryptic so long as the `html_test.exs` tests capture that clearly.

germsvel (commented):
Love the work! 🙌 

Thanks for doing this. I left a few comments. The biggest one (I think) is what should we call the option? Otherwise, this is looking great!

lib/phoenix_test/assertions.ex
ftes: `selected_value` is my least favourite, because it might imply you're asserting on the selected option's `value` rather than the option's label.

How about `selected`. It's not precise (as in `selected_option_label`), but given PhoenixTest preference for operating on visible labels, I think this is consistent and concise.

germsvel: Yeah, I'm with you. I think `selected` is the best of the three.

lib/phoenix_test/html.ex
ftes: Made the code more explicit and added tests.

lib/phoenix_test/assertions.ex
```diff
@@ -435,53 +441,58 @@ defmodule PhoenixTest.Assertions do
...
+    content =
+      opts
+      |> Map.take(~w(text value selected)a)
+      |> Enum.reject(&(&1 in [text: :no_text, value: :no_value, selected: :no_selected]))
 
-  defp finder_fun(selector, %Opts{text: :no_text, value: :no_value, label: label} = opts, _operation)
-       when is_binary(label) do
-    &Query.find_by_label(&1, selector, label, Opts.to_list(opts))
-  end
+    case {content, opts, operation} do
```
ftes, line 459: I tried to make the `conflicting content opts` more explicit.

That led me to discover that `text + label` opts are unsupported (label silently ignored). I made that explicit by raising an error instead.

lib/phoenix_test/assertions.ex
```diff
@@ -435,53 +441,58 @@ defmodule PhoenixTest.Assertions do
...
+      {[], %Opts{label: :no_label}, _} ->
+        &Query.find(&1, selector, Opts.to_list(opts))
 
-  defp finder_fun(selector, %Opts{text: :no_text, value: value} = opts, _operation) do
-    value_finder_fun(ensure_binary(value), selector, opts)
-  end
+      {[], %Opts{label: label}, _} when is_binary(label) ->
```
ftes, line 463: We keep repeating the `is_binary(label)` guard and `ensure_binary(value)` type coercions.

That could be hoisted into `Opts.parse` - since we don't support different Opts types elsewhere.

Didn't want to change too much here though.

germsvel (approved):
Lovely! Thanks @ftes!

PR #310, opened 2026-04-23
Refactor: Revert changes from #309 after introducing separate `selected` option

ftes:
Follow up for #309.
With the new approach (separate `selected` option), some changes to existing `value` finder path became unnecessary.
Sorry, I realized this a little too late.

Take this or leave it - #309 was ok - just a bit more bloated than it had to be.


germsvel:
> Sorry, I realized this a little too late.

No worries. I appreciate the follow-up. And I'll take it! Less code is better!

PR #311, opened 2026-04-23
Support async navigation in assert/refute_path

### LINKED ISSUE #209: Handle current_path patch via LiveViewWatcher
If the LiveView path is patched by an async event (e.g. pubsub) it seems that `%Live{current_path: _}` is not updated.
So even if one manually awaits the pubsub event that triggers the patch, a subsequent `assert_path` fails.

Should `assert_path` be wrapped with the same timeout mechanism used for `assert_has` and `refute_has`?

ftes:
Close #209 

Add `timeout` option to `assert_path` and `refute_path`, so they can react to async navigation.

### DIFF
--- a/lib/phoenix_test/assertions.ex
+++ b/lib/phoenix_test/assertions.ex
@@ -270,29 +270,33 @@ defmodule PhoenixTest.Assertions do
     session
   end
 
-  def assert_path(session, path, opts \\ [])
-
-  def assert_path(session, path, []) do
+  def assert_path(session, path, opts \\ []) do
     session = set_operation(session, :assert_path, "")
     uri = URI.parse(PhoenixTest.Driver.current_path(session))
+    params = opts[:query_params] && Utils.stringify_keys_and_values(opts[:query_params])
 
-    if path_matches?(path, uri.path) do
-      assert true
-    else
-      msg = """
-      Expected path to be #{inspect(path)} but got #{inspect(uri.path)}
-      """
+    cond do
+      not path_matches?(path, uri.path) ->
+        msg = """
+        Expected path to be #{inspect(path)} but got #{inspect(uri.path)}
+        """
 
-      raise AssertionError, msg
+        raise AssertionError, msg
+
+      is_nil(params) or query_params_match?(session, params) ->
+        assert true
+
+      true ->
+        params_string = Plug.Conn.Query.encode(params)
+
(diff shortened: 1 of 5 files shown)


lib/phoenix_test/assertions.ex
```diff
@@ -288,11 +288,15 @@ defmodule PhoenixTest.Assertions do
...
-    |> assert_path(path)
-    |> assert_query_params(params)
+    case Keyword.fetch(opts, :query_params) do
+      {:ok, params} ->
+        session
+        |> assert_path(path)
+        |> assert_query_params(params)
+
+      :error ->
+        assert_path(session, path)
```
germsvel: Out of curiosity, why this change? Might be obvious, and I'm just missing it.

germsvel (commented):
I think this looks mostly good. Just a minor comment (might be my brain is tired)

lib/phoenix_test/assertions.ex
ftes: No, it's not obvious, agreed.
I think the `Keyword.fetch + :error` doesn't convey the intent well.

`query_params` was the only supported opt, and it was required.
Calling `assert_path/3` without `query_params` resulted in errors.
Now we always call `assert_path` with opts, sometimes `opts is []` empty list.

I made this easier to read in the first commit.
And added a second, optional, refactoring commit 58a9ce1e6836e2478ee8971806f2a2697a605cb7 to this PR.

lib/phoenix_test/assertions.ex
```diff
@@ -317,73 +321,40 @@ defmodule PhoenixTest.Assertions do
...
-  defp refute_query_params(session, params) do
-    params = Utils.stringify_keys_and_values(params)
+        raise AssertionError, msg
 
-    uri = URI.parse(PhoenixTest.Driver.current_path(session))
-    query_params = uri.query && URI.decode_query(uri.query)
+      query_params_match?(session, params) ->
+        params_string = URI.encode_query(params)
 
-    if query_params == params do
```
ftes, line 377: We used to handle `(is_nil(query_params) and params == %{})` in `assert_path`, but not in `refute_path`.

By extracting a common helper, the behaviour is now symmetrical.

germsvel (approved):
Love it. Thanks for the work @ftes!

PR #313, opened 2026-04-26
Add :checked option to assert_has/3 and refute_has/3

### LINKED ISSUE #213: Feature request: `assert_checked/{2,3}` and `refute_checked/{2,3}`
Hi German!

I've been testing a bunch of complicated forms recently, where changing one control can affect the state of checkboxes elsewhere in the form. It'd be nice to have a `assert_checked`/`refute_checked` functions that use the same logic as `check/{3,4}` for going from the label of a checkbox to the checkbox input itself. The current workaround requires grabbing the ID or input name of the checkbox, which seems counter to the philosophy of the library.

I'm happy to submit a PR if this is something you'd like. ☺️

s3cur3:
Adds support for assertions like:

```elixir
assert_has(session, ".user", checked: true, label: "Frodo")
assert_has(session, ".user", checked: false, label: "Frodo")
```

Resolves #213

### DIFF
--- a/lib/phoenix_test.ex
+++ b/lib/phoenix_test.ex
@@ -1343,11 +1343,12 @@ defmodule PhoenixTest do
   than one matching element is found (unless `:count` option is used).
 
   You can assert an element's contents using `:text`, assert a field's value
-  using `:value` (with an optional `:label`), or assert a `select` element's
-  selected option text using `:selected`.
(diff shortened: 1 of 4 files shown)


germsvel (approved):
Love it! Thanks for this work @s3cur3!

PR #314, opened 2026-04-29
Fix: Omit hidden input before checkbox group

ftes:
Omit hidden 'reset' input that appears before checkbox group if any checkboxes are checked.

## What changed?
- Better preserve DOM order of form elements better when building form data.

## Why make these changes?
A hidden input before a checkbox group can be used to reset the value if all checkboxes are unchecked.
However, if checkboxes are checked, then the hidden input should be effectively ignored.
Plug/Phoenix can handle this if the field values arrive in DOM order.

In forms like:

```html
<input type="hidden" name="list" value="" />
<input type="checkbox" name="list[]" value="one" />
<input type="checkbox" name="list[]" value="two" />
```

PhoenixTest previously collected controls by selector category and stored `FormData` in a map. That lost DOM order and collapsed repeated field names, which could cause the hidden scalar entry to interfere with the checkbox array payload.

This PR is only a partial improvement. The ideal solution would be to submit form data in the exact DOM order a browser would, end to end.

### DIFF
--- a/lib/phoenix_test/element/form.ex
+++ b/lib/phoenix_test/element/form.ex
@@ -69,7 +69,7 @@ defmodule PhoenixTest.Element.Form do
 
   def has_action?(form), do: Utils.present?(form.action)
 
-  @enabled_controls "input:not([disabled])[name], textarea:not([disabled])[name], select:not([disabled])[name]"
+  @enabled_controls ":is(input, textarea, select):not([disabled])[name]"
 
   defp form_data(form) do
     form
--- a/lib/phoenix_test/element/form.ex
+++ b/lib/phoenix_test/element/form.ex
@@ -69,70 +69,58 @@ defmodule PhoenixTest.Element.Form do
 
   def has_action?(form), do: Utils.present?(form.action)
 
-  @simple_value_types ~w(
-    date
-    datetime-local
-    email
-    month
-    number
-    password
-    range
-    search
-    tel
-    text
-    time
-    url
-    week
-  )
-
-  @hidden_inputs "input[type='hidden']"
-  @checked_radio_buttons "input:not([disabled])[type='radio'][value]:checked"
-  @checked_checkboxes "input:not([disabled])[type='checkbox'][value]:checked"
-  @pre_filled_default_text_inputs "input:not([disabled]):not([type])[value]"
-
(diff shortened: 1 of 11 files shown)


lib/phoenix_test/element/form.ex
```diff
@@ -69,70 +69,58 @@ defmodule PhoenixTest.Element.Form do
...
-  @checked_radio_buttons "input:not([disabled])[type='radio'][value]:checked"
-  @checked_checkboxes "input:not([disabled])[type='checkbox'][value]:checked"
-  @pre_filled_default_text_inputs "input:not([disabled]):not([type])[value]"
-
-  @pre_filled_simple_value_inputs Enum.map_join(
-                                    @simple_value_types,
-                                    ",",
-                                    &"input:not([disabled])[type='#{&1}'][value]"
-                                  )
+  @enabled_controls "input:not([disabled])[name], textarea:not([disabled])[name], select:not([disabled])[name]"
```
ftes:
```suggestion
  @enabled_controls ":is(input, textarea, select):not([disabled])[name]"
```

lib/phoenix_test/form_data.ex
```diff
@@ -35,82 +35,106 @@ defmodule PhoenixTest.FormData do
...
-      {_key, _value} = field ->
-        field
-    end)
-    |> List.flatten()
+  defp replace_entries(data, name, new_entries) do
+    case Enum.find_index(data, fn {n, _} -> n == name end) do
+      nil ->
+        data ++ new_entries
+
+      first_index ->
```
ftes, line 128: This is a bit of a crux to handle multiple fields with same name.

germsvel:
Thanks for the work @ftes! I was really hoping we wouldn't have to revert back to lists for `FormData`, but it makes sense that we want to preserve that order. I can't recall now how we were preserving it before. 

To your comment: 

> The ideal solution would be to submit form data in the exact DOM order a browser would, end to end.

I've been wondering if there's a way to do that. In LiveViews, we might not need to keep track of FormData. In dead views, maybe? But even then, I wonder if when we submit the form, we don't use `FormData`, but rather whatever is on the page itself. 

In any case, this is fixing an issue, and doesn't seem to break other things, so I'm gonna go ahead and merge it.

PR #317, opened 2026-05-21
Bugfix: nested within was resetting to top-level

markevans:
Hi - thanks for this useful library.

This is a fix for things like the following:

```html
<div class="page-1">
  <p>...</p>
  <button>Next</button>
</div>

<div class="page-2">
  <button>Next</button>
</div>

```

```elixir
conn
|> within(".page-1", fn page ->
  |> within("p", fn p ->
    # assert something...
  end)
  # THIS BREAKS! due to finding multiple buttons because `within` resets scope to :none
  |> click_button("Next")
end)
```

`within` was resetting the scope to `:none`, rather than the parent scope.


germsvel (approved):
Thanks for the work @markevans!

I don't think nesting `within`s is a great idea. At least, it never occurred to me that we would want to nest things like that. It probably makes tests more confusing. But I don't see the harm in supporting this.

markevans:
That's fair - the reason I wanted it was for working with a page that had multiple similar tables, where I wanted to e.g. click a button within a specific table row in a specific table - but indeed maybe there's a better way I could have used without multiple nesting.
Anyway thanks for this!

PR #319, opened 2026-07-15
feat: Support finding elements by aria-label and aria-labelledby (form helpers + assertions)

### LINKED ISSUE #243: Find by aria-label
It would be great to be able to find elements by its `aria-label` property, for example, this input should be findable by its `aria-label` "Your Name"

```html
<input id="name" type="text" name="name" aria-label="Your Name"/>
```

I think that's not possible currently. Are there plans to support ARIA attributes?

Wigny:
Closes #243.

### DIFF
--- a/lib/phoenix_test.ex
+++ b/lib/phoenix_test.ex
@@ -588,6 +588,13 @@ defmodule PhoenixTest do
   <label for="search" class="sr-only">Search</label>
   <input id="search" type="text" name="q" />
   ```
+
+  In addition to `<label>` elements, the label can also be the accessible name
+  provided by the input's `aria-label` or `aria-labelledby` attribute:
+
+  ```html
+  <input type="text" name="q" aria-label="Search" />
+  ```
   """
   def fill_in(session, label, opts) when is_binary(label) and is_list(opts) do
     opts = Keyword.validate!(opts, [:with, exact: true])
@@ -1367,7 +1374,11 @@ defmodule PhoenixTest do
     Pass `true` to look for a checked field, or `false` to look for an
     unchecked field.
 
-  - `label`: the label associated to the form field with `value`, `selected`, or `checked`
+  - `label`: the label associated to the form field with `value`, `selected`, or
+  `checked`. The association can be a `<label>` element (wrapping the field or
+  pointing to it via `for`/`id`), or an accessible name provided by the field's
+  `aria-label` or `aria-labelledby` attribute. Note: if a field is reachable by
+  both a `<label>` and an `aria-*` attribute, the `<label>` match takes precedence.
 
   - `exact`: by default `assert_has/3` will perform a substring match (e.g. `a
   =~ b`). That makes it easier to assert text within HTML elements that also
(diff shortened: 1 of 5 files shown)


lib/phoenix_test.ex
```diff
@@ -1367,7 +1374,11 @@ defmodule PhoenixTest do
     Pass `true` to look for a checked field, or `false` to look for an
     unchecked field.
 
-  - `label`: the label associated to the form field with `value`, `selected`, or `checked`
+  - `label`: the label associated to the form field with `value`, `selected`, or
+  `checked`. The association can be a `<label>` element (wrapping the field or
+  pointing to it via `for`/`id`), or an accessible name provided by the field's
+  `aria-label` or `aria-labelledby` attribute. Note: if a field is reachable by
+  both a `<label>` and an `aria-*` attribute, the `<label>` match takes precedence.
```
germsvel, line 1381: Thanks for adding docs! 

I know this is true in the code, but I wonder if we should do it the other way around: `aria-labelled-by` > `aria-label` > `label/inner-text`. I believe that's how browsers sort the priority (I was reading [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label)). 

But that also brings up a problem -- finding by `aria-labelled-by` is expensive, since we're trying to find it across all the DOM :( So, maybe this is a good trade-off? I don't know, to be honest. I'm gonna think about it a little.

germsvel (commented):
Thanks for opening this PR @Wigny! I really appreciate it. 

(sorry for radio silence on my side. I get back to `phoenix_test` as much as I can, but sometimes I'm too swamped). 

There are 2 open questions that I need to think about before merging this: 

1. Should we change the order of precedence? `aria-labelled-by` > `aria-label` > `label`/`inner text`? I believe that's how the browser works. We've tried to keep in line with that as much as possible, but it might be overly expensive here, and it might be worth the trade-off to do it like you have here. That's been an issue I've run into before when thinking about implementing this, so perhaps it just makes sense to support it in a different order. 

2. This only supports finding elements by `aria-label`/`aria-labelled-by` in assertions. Things like `click_button("Close")` where "Close" is the `aria-label="Close"` don't work with these changes (I tested that briefly). And I'm not sure if this supports form helpers (like `fill_in`, `check`, etc.). That's not the end of the world. It just means we can't close the related GitHub issue. But we could get this in, and then try to layer on support for aria labels in a separate PR. 

What do you think? You interested in taking a stab at support `aria-label` and labelled by for other things too? Or do you want to change the title/description to be about aria-label in assertion helpers?

germsvel:
Actually, since this is a net improvement. I'm okay with going with this. I also think it's okay that the precedence is inverted, especially since you mentioned it in the docs. 

I'm gonna go ahead and merge this as is, and then, I'll see about layering on `aria-label` support for `click_button`/`click_link` (and form helpers, if they need it too)

But if you have thoughts on the precedence part, I'd love to hear them! We can always change things a bit.

